Issue #9, opened 2020-09-04
Provide transform/function to get coordinates in world space


aevyrie:
Coordinates are currently reported in ndc space. Use the associated camera transform to convert to world space, evaluating lazily.

aevyrie:
PR #14 implements a function to translate from ndc to world space. To resolve this issue, there should be a function that uses the camera associated with the current PickableMesh to supply the transforms into this function automatically.

guimcaballero:
Another thing to note, the current implementation from #14 considers that the ndc coordinates are correct, so due to #5 world space coords might be wrong for objects such as large planes, where the depth from the first vertex is far away from where the intersection is.

aevyrie:
As of #26 default for everything is world space, closing.

PR #14, opened 2020-09-05
Implement method to get world coordinates in PickIntersection

guimcaballero:
I started some work on #9. It's a simple function that transforms the ndc coordinate into world coordinate, and takes in a view and projection matrix. 

I'm not sure if this is the best way to implement it, I'm open to ideas. I just needed something to get going :P.

As a side note, here's what I'm using it for:

bavy

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,10 +48,23 @@ impl Default for PickState {
 pub struct PickIntersection {
     entity: Entity,
     pick_coord_ndc: Vec3,
+    pick_world_pos: Vec3,
 }
 impl PickIntersection {
-    fn new(entity: Entity, pick_coord_ndc: Vec3) -> Self {
-        PickIntersection { entity, pick_coord_ndc }
+    fn new(entity: Entity, pick_coord_ndc: Vec3, pick_world_pos: Vec3) -> Self {
+        PickIntersection {
+            entity,
+            pick_coord_ndc,
+            pick_world_pos,
+        }
+    }
+
+    pub fn get_pick_world_pos(&self) -> Vec3 {
+        self.pick_world_pos
+    }
+
+    pub fn get_pick_coord_ndc(&self) -> Vec3 {
+        self.pick_coord_ndc
     }
 }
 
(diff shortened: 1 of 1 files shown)


aevyrie:
Awesome, thank you! I'll have to take a look when I get some time, it's currently a holiday weekend for me. 

I'd like to build on this to add a 3d cursor (small sphere) in world space to help with debug, this might help us figure out the plane issue you found, because we can visualize where those world space intersections are being calculated. I know they are incorrect right now, and it will snap to vertices.

Anyway, that can come later. I'll merge this as soon as I can take a look.

guimcaballero:
Sure thing, no rush! I just wanted to get started with a basic movement system, so I'm making the features I need on the way.

I'm currently building my game with a dependency to my fork, so I can keep working on it while this pr isn't merged.

The 3d cursor seems a great idea for debugging, although I'm not sure what's the best approach. Maybe a system that moves the sphere to the `pick_state.top()` pick world position?

aevyrie:
> Maybe a system that moves the sphere to the pick_state.top() pick world position?

Yep, I think so.

aevyrie (changes requested):
This looks good, thank you!

My only request is that you change the name from `get_pick_world_pos` to `get_pick_coord_world`. I haven't planned out all the function names or anything like that, but I'd like to keep them consistent where possible. This would then match the existing function `get_pick_coord_ndc`.

So you're aware, I'd eventually like to expose this to the user through the PickableMesh component. To guarantee the output world coordinates are correct, this function must be supplied a correct view and projection matrix. The `bevy_mod_picking` plugin already has that information when the user associates a camera with the PickableMesh components, so I should probably figure out a way to use that. 😄

guimcaballero:
Yeah, makes more sense as `get_pick_coord_world`. I've also moved it after `get_pick_coord_ndc`.

And the PickableMesh things sounds good!

aevyrie:
Great, thanks!

aevyrie (approved):

Issue #15, opened 2020-09-05
Sometimes pick_state.top() returns None for large Planes


guimcaballero:
Something weird happens sometimes, where a `pick_state.top()` returns None even if there is a Plane there. The issue occurs when the Plane is large, like size `50.0`. I don't have the problem with size `10.0`.

The Plane is created like this:

```rust
    commands.PbrComponents {
            mesh: meshes.add(Mesh::from(shape::Plane { size: 50.0 })),
            material: materials.add(Color::rgb_u8(222, 184, 135).into()),
            ..Default::default()
        })
        .with(PickableMesh::new(camera_entity));
```

I've recorded a couple gifs of this happening, with the same system used in the example:

```rust
fn get_picks(pick_state: ResMut<PickState>) {
    println!("All entities:\n{:?}", pick_state.list());
    println!("Top entity:\n{:?}", pick_state.top());
}
```

![bavy_issue1](https://user-images.githubusercontent.com/29580620/92306046-a8857580-ef8c-11ea-9d16-a25b60bba796.gif)
![bavy_issue2](https://user-images.githubusercontent.com/29580620/92306052-ae7b5680-ef8c-11ea-8a28-5ff6a351f507.gif)

guimcaballero:
It seems like the example completely breaks when the Plane size is set to `50.0`. 

![bavy_issue3](https://user-images.githubusercontent.com/29580620/92306410-2fd3e880-ef8f-11ea-94c1-45bf0e686cf4.gif)

The plane is only highlighted when the mouse goes outside. 

Any ideas why this might be happening?

aevyrie:
Thanks for reporting this! I'm almost certain this is related to #5 and #6 . I'd have to do some debugging with the triangle area calculation code however. I wonder, does this happen to other objects that are partially off screen?

aevyrie:
The vertices being used by the function look a bit odd. The first two numbers are tri area, and cursor tri area (hit if equal), and the three pairs that follow are the x/y coordinates of the vertices in ndc.,t his is with a plane size of 18:
No Hit: 493.499  743.078  0.359,0.710  **70.805,139.869**  2.939,-1.200 
Hit: 182.905  182.905  0.359,0.710  -1.089,0.445  **70.805,139.869**

After plotting the points, they don't make a ton of sense compared to what I see on screen.
After dropping the plane size to 17, picking behaves as expected. Points look like this:
[...]

PR #16, opened 2020-09-06
Change point_in_tri to use barycentric coordinates

guimcaballero:
I'm opening this PR because I have the function already written for debugging #15. 

It's currently pretty much untested, I just played around with the example a bit and checked that it works with the Plane set to `10.0`. I'll test it today/tomorrow to make sure it works at least as well as the current implementation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -419,21 +419,21 @@ fn pick_mesh(
     }
 }
 
-/// Compute the area of a triangle given 2D vertex coordinates, "/2" removed to save an operation
-fn double_tri_area(a: &Vec2, b: &Vec2, c: &Vec2) -> f32 {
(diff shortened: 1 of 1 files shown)


aevyrie:
Thank you!

> I'll test it today/tomorrow to make sure it works at least as well as the current implementation.

This brings up a good point. Do you you have any ideas how we might develop integration/unit tests to prevent regressions to some of these fixes? It may require breaking out most logic to testable functions, and the bevy systems would be limited to getting data from queries/resources, and passing the data into these functions. That would allow us to emulate user input in unit tests without diving into the bevy backend. I'm completely inexperienced in this area, and I'd love any input.

aevyrie (approved):

PR #22, opened 2020-09-08
Add 3d debug cursor and DebugPlugin

guimcaballero:
As I mentioned in #21, I've created the 3d debug cursor discussed in #14. 

I've decided to create a plugin called `DebugPickingPlugin` which sets everything up, where future debug tools can be added. This way all the debug stuff can be added easily and independently from the main plugin. I think this is the cleanest approach, what do you think?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,6 +18,14 @@ impl Plugin for PickingPlugin {
     }
 }
 
+pub struct DebugPickingPlugin;
+impl Plugin for DebugPickingPlugin {
+    fn build(&self, app: &mut AppBuilder) {
+        app.add_startup_system(setup_debug_cursor.system())
+            .add_system(update_debug_cursor_position.system());
+    }
+}
+
 pub struct PickState {
     cursor_event_reader: EventReader<CursorMoved>,
     ordered_pick_list: Vec<PickIntersection>,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -195,6 +203,54 @@ struct NdcBoundingCircle {
...
+        view_matrix = transform.value.inverse();
+        projection_matrix = camera.projection_matrix;
+    }
+
+    // Set the cursor translation to the top pick's world coordinates
+    if let Some(top_pick) = pick_state.top() {
+        let pos = top_pick.get_pick_coord_world(projection_matrix, view_matrix);
+
+        for (_, mut translation) in &mut query.iter() {
+            translation.0.set_x(pos.x());
```
aevyrie, line 227: Can
`translation.0 = pos`
be used instead of passing individual axes?

aevyrie (commented):
Small note about the cursor position update. Thanks for taking this on, love the idea to use a system, I agree it's a clean way to accomplish this. 😁

src/lib.rs
guimcaballero: Oops, yeah. Fixed!

aevyrie (approved):
Nice!

Issue #23, opened 2020-09-09
bevy_mod_picking doesn't compile on stable


bitshifter:
When running `cargo build --examples` I get the error:

```sh
error: failed to run `rustc` to learn about target-specific information

Caused by:
  process didn't exit successfully: `rustc - --crate-name ___ --print=file-names -Clink-arg=-fuse-ld=lld -Zshare-generics=y --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg` (exit code: 1)
  --- stderr
  error: the option `Z` is only accepted on the nightly compiler
```

This is because `.cargo/config` is in .git and contains options which are only available on nightly.

I think `.cargo/config` shouldn't be committed. Note that `bevy` itself has this file as `.cargo/config_fast_builds` so it won't actually be used until it's been renamed.

aevyrie:
Thanks for reporting this - I added .cargo/config to the .gitignore, but I clearly need to rename it as you suggested.

aevyrie:
Latest PR has fixed this. Thanks @bitshifter!

PR #25, opened 2020-09-10
Fix cargo config breaking stable compile

aevyrie:
Fixing issue #23


aevyrie:
Tested working, merging now!

PR #26, opened 2020-09-10
World space ray casting

aevyrie:
This is a draft of world space ray casting. This detects cursor picking by casting a ray in world space against mesh triangles in world space. Once implemented, this should resolve a number of current issues. The `ray_triangle_intersection` function returns `None` if there is no intersection, and returns `Some` containing the hit position and normal direction if there is an intersection.

I've also moved the ray casting logic and structs into a new file `raycast.rs`. This should make testing and benchmarking much easier, because we can pass inputs directly to these functions independent of any Bevy systems or resources.

@guimcaballero This will replace much/all of the ndc related code, and hopefully simplify the whole plugin. I want you to be aware of this so you don't waste any effort on something that will be replaced. 🙂

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -342,6 +342,30 @@ fn select_mesh(
     }
 }
 
+fn pick_mesh_raycast(
+     // Resources
+     mut pick_state: ResMut<PickState>,
+     cursor: Res<Events<CursorMoved>>,
+     meshes: Res<Assets<Mesh>>,
+     windows: Res<Windows>,
+     // Queries
+     mut mesh_query: Query<(&Handle<Mesh>, &Transform, &mut PickableMesh, Entity)>,
+     mut camera_query: Query<(&Transform, &Camera)>,
+) {
+    // Get the cursor position
+    let cursor_pos_screen: Vec2 = match pick_state.cursor_event_reader.latest(&cursor) {
+        Some(cursor_moved) => cursor_moved.position,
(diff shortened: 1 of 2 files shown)


guimcaballero:
Thanks for the heads up, sounds really promising!

aevyrie:
Somewhat shockingly, it's already in a semi working state. I'm getting hits detected when I move the mouse around the screen, but the results aren't correct. WIll have to spend some time debugging; that said, I wasn't expecting to get this far so quickly.

aevyrie:
It looks like the pick ray construction is working as intended. The debug cursor follows the mouse around, which tells me the ray is projecting from the camera correctly. Most areas of the screen are dead, but a few spots that should be empty are pickable. I think this implies the error is somewhere in the mesh transformation or triangle intersection logic.

aevyrie:
@guimcaballero So I've got it working* after implementing the Moller-Trumbore algorithm, but I've now got the strangest bug I've ever seen. Each time I compile and run, a new combination of meshes is no longer selectable. 😖 

ezgif-2-577d0e32a97d

aevyrie:
Everything fixed up, seems to be working as intended. 😄 

ezgif-2-c2c155425683

guimcaballero:
Looks awesome!

aevyrie:
Just did a rebase, sounds like you're happy with this, so I'll merge.

PR #27, opened 2020-09-14
Add Entity getter to PickIntersection

guimcaballero:
After #26, I'm opening two PRs with some small changes I've been trying on my [develop](https://github.com/guimcaballero/bevy_mod_picking/tree/develop) branch.

This one adds simply an entity getter to PickIntersection. Just a question about naming, should it be `get_entity()` or just `entity()`?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -67,6 +67,10 @@ impl PickIntersection {
             distance,
         }
     }
+    /// Entity intersected with
+    pub fn entity(&self) -> Entity {
+        self.entity
+    }
     /// Position vector describing the intersection position.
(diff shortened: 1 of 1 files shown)


aevyrie (approved):

aevyrie:
@guimcaballero `entity()` is correct, if we're following the Rust API guidelines: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

Only thing I wasn't sure about in this PR is returning a reference to the Entity. The existing functions position() and normal() return a reference to a Vec3. I think your solution is correct though, Entity is just a wrapper around a u128, and it derives Copy.

PR #29, opened 2020-09-15
Update transforms after rework in bevy#374

guimcaballero:
PR [#374](https://github.com/bevyengine/bevy/pull/374) changed how transforms work on bevy. This PR updates everything to be up to date.

Currently there's an issue I haven't been able to fix:

bevy

The cursor jumps around when on the ball, but not on the cube or the plane. 

I'll try to fix it tomorrow, but if you see what I'm doing wrong please let me know.

### DIFF
--- a/examples/3d_scene.rs
+++ b/examples/3d_scene.rs
@@ -35,7 +35,7 @@ fn setup(
         .spawn(PbrComponents {
             mesh: meshes.add(Mesh::from(shape::Cube { size: 1.0 })),
             material: materials.add(Color::rgb(1.0, 1.0, 1.0).into()),
-            translation: Translation::new(0.0, 1.0, 0.0),
+            transform: Transform::from_translation(Vec3::new(0.0, 1.0, 0.0)),
             ..Default::default()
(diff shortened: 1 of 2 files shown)


aevyrie:
Thanks for the heads up, I wasn't aware of this upcoming change! That's an odd bug, I'll try and take a look at your code today.

aevyrie:
Found the issue. 😄 

Because the translation and rotation updates were happening sequentially, It looks like the screen was occasionally refreshing between these updates, causing the debug cursor to fly off in space because the translation had update but not the rotation.

Changing the debug cursor update to this fixed the problem:
```rust
let transform_new = Mat4::from_rotation_translation(new_rotation, *position);
for mut transform in &mut query.iter() {
    *transform.value_mut() = transform_new;
}
```

aevyrie:
Merging this ASAP because current version won't compile with the bevy update. Thanks @guimcaballero!

guimcaballero:
Oh wow, didn't expect that. Thanks for the help!

PR #35, opened 2020-09-19
Update for Entity and Query API changes

UnFlimFlammable:


examples/3d_scene.rs
```diff
@@ -20,10 +20,22 @@ fn setup(
...
+        .spawn(Camera3dComponents {
+            transform: Transform::new(Mat4::face_toward(
+                Vec3::new(-3.0, 5.0, 8.0),
+                Vec3::new(0.0, 0.0, 0.0),
+                Vec3::new(0.0, 1.0, 0.0),
+            )),
+            ..Default::default()
+        });
+
+    let camera_entity = commands.current_entity().unwrap();
```
aevyrie, line 35: I think you can simplify this block by changing it to
```rust
let camera_entity = commands
    .spawn(nonsense here)
    .current_entity().unwrap();
```
This will also reduce the chance of someone (me) adding another spawn block before the `camera_entity` definition, which would make it point to some entity other than the camera.

examples/stress_test.rs
```diff
@@ -22,11 +22,30 @@ fn setup(
...
+                edge_length as f32 * 0.8,
+                edge_length as f32 * 1.2,
+            ),
+            Vec3::new(0.0, -(edge_length as f32) * 0.1, 0.0),
+            Vec3::new(0.0, 1.0, 0.0),
+        )),
+        ..Default::default()
+    });
+
+    let camera_entity = commands.current_entity().unwrap();
```
aevyrie, line 47: Same thing here as above
let camera_entity = commands.spawn()....

aevyrie (changes requested):

examples/3d_scene.rs
UnFlimFlammable: Updated

examples/stress_test.rs
UnFlimFlammable: Updated

aevyrie:
Only issue is rustfmt, no biggie, want to get this merged asap. Thanks for this work!

PR #36, opened 2020-09-19
use camera entity for picking

verzuz:
this should allow to use more than one camera

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,7 @@ use bevy::{
     window::CursorMoved,
 };
 use raycast::*;
+use std::collections::HashMap;
 
 pub struct PickingPlugin;
 impl Plugin for PickingPlugin {
@@ -278,7 +279,7 @@ fn pick_highlighting(
         &SelectablePickMesh,
         &Handle<StandardMaterial>,
     )>,
-    query_selectables: Query<&SelectablePickMesh>,
+    mut query_selectables: Query<&SelectablePickMesh>,
 ) {
     // Query selectable entities that have changed
     for (mut highlightable, selectable, material_handle) in &mut query_selected.iter() {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -374,94 +375,100 @@ fn pick_mesh(
...
-    // Get the view transform and projection matrix from the camera
-    let mut camera_matrix = Mat4::zero();
-    let mut projection_matrix = Mat4::zero();
-    for (transform, camera) in &mut camera_query.iter() {
-        camera_matrix = *transform.value();
-        projection_matrix = camera.projection_matrix;
-    }
-    let (_, _, camera_position) = camera_matrix.to_scale_rotation_translation();
+    // collect and calculate pick_ray and camera_position from all cameras
+    let mut rays: HashMap<Entity, (Ray3D, Vec3)> = HashMap::new();
```
aevyrie, line 379: The camera position Vec3 here is redundant - the ray origin is the camera position. You can simplify this to a Hashmap of type <Entity, Ray3D>

verzuz:
recheck

aevyrie (commented):
Thank you for submitting this! I really only see one thing I'd like to change, and one thing I want to have a discussion about.

I left a comment, but the camera position in the hashmap tuple is redundant - that is already stored inside the Ray3d.

That leads to a larger question though - is the hashmap actually necessary? This only seems to make sense if there can be picks coming from multiple cameras at the same time. I would think that for a picking system, there will only ever be one active camera/window at a given time.

I'd like to hear your thoughts!

aevyrie:
Thinking more about this, there is a lot that may need to change in order to allow for multi-window support as well as using a color picking shader. Instead of agonizing over it now, I think I should maybe just merge this as is, so we can play with it, while I continue to design the system for handling multi-window etc.

verzuz:
Right now it at least allows to have multiple cameras even if we only use one for picking - this will allow us to use this plugin with ie the UIPlugin that brings it's own camera. I'm curious if we can render and pick different entities with different cameras with this change. I'll try to setup an example next week.

PR #42, opened 2020-09-21
Experimental Pick Groups and Custom Rays

aevyrie:
This branch is for developing "pick groups".

The intent here is to:

1. Allow the definition of sets of picking groups which relate meshes with cameras (and eventually windows).
2. Separate ray construction from picking logic, allowing the user to feed in arbitrary picking rays, not necessarily originating from the cursor. See #43.


nickbryan:
Hey, I tried this out and the API seems really nice but I am getting an error stating `thread 'Compute Task Pool (3)' panicked at 'Multiple PickingSources have been added to pick group: 0'`.

Apologies for the long code snippet but I thought that it may add context as to how I am using it. I am not sure if I am misunderstanding the use of Center but I have tried with `PickingMethod::ScreenSpace(Vec2::zero())` too and getting the same message. I have also tried with different group numbers (other than `0`). I have removed some code for brevity (although it is not brief) and added some comments to mark where the picking code is used.
```rust
fn main() {
    App::build()
        .add_default_plugins()
        .add_plugin(camera::CameraPlugin)
        .add_plugin(PickingPlugin)
        .add_startup_stage_after(camera::STARTUP_STAGE, "main")
        .add_startup_system_to_stage("main", setup.system())
        .run();
}

fn setup(
    mut commands: Commands,
    camera_pick_group: Res<camera::CameraPickingGroup>,
    mut meshes: ResMut<Assets<Mesh>>,
) {
    let my_meshes = ...;

    for mesh in my_meshes {
        commands
            .spawn(PbrComponents {
                ...
            })
            .with(PickableMesh::new([camera_pick_group.0].into())); // Use the picking group here <------
    }
}

mod camera {
    pub const STARTUP_STAGE: &str = "camera_startup_stage";

    pub struct CameraPlugin;

    impl Plugin for CameraPlugin {
        fn build(&self, app: &mut AppBuilder) {
            app.add_startup_stage(STARTUP_STAGE)
                .add_startup_system_to_stage(STARTUP_STAGE, setup.system())
                .init_resource::<RotateSystemState>()
                .add_system(rotate_system.system());
        }
    }

    pub struct CameraPickingGroup(pub PickingGroup); // Used to track this particular group <-------

    struct CameraComponent;

    fn setup(mut commands: Commands) {
        let picking_group = PickingGroup::Group(0); // PICKING GROUP CREATED HERE <-------

        commands
            .spawn(Camera3dComponents {
                transform: Transform::from_translation_rotation(
                    Vec3::new(25.0, 20.0, 25.0),
                    Quat::from_rotation_ypr(
                        STARTING_YAW.to_radians(),
                        STARTING_PITCH.to_radians(),
                        0.0,
                    ),
                ),
                ..Default::default()
            })
            .with(CameraComponent)
            .with(PickingSource::new(
                picking_group,
                PickingMethod::Center,
            ));

        commands.insert_resource(CameraPickingGroup(picking_group)); // Added to a resource so we can use it in another system where the meshes are defined <-------
    }

    #[derive(Default)]
    struct RotateSystemState {
        mouse_motion_event_reader: EventReader<mouse::MouseMotion>,
        focus: Option<Vec3>,
    }

    fn rotate_system(
        mut state: ResMut<RotateSystemState>,
        keyboard_input: Res<Input<KeyCode>>,
        mouse_button_input: Res<Input<MouseButton>>,
        events: Res<Events<mouse::MouseMotion>>,
        pick_state: Res<bevy_mod_picking::PickState>, // Pick state <-------
        pick_group: Res<CameraPickingGroup>, // Get the picking group  <-------
        windows: Res<Windows>,
        mut camera_query: Query<(&CameraComponent, &mut Transform)>,
    ) {
        if keyboard_input.pressed(KeyCode::LShift) && mouse_button_input.pressed(MouseButton::Left) {
            let mut rotation_move = Vec2::zero();

            for event in state.mouse_motion_event_reader.iter(&events) {
                rotation_move -= event.delta;
            }

            match state.focus {
                Some(focus) => {
                    let window = windows.get_primary().unwrap();
                    let screen_width = window.width as f32;
                    let screen_height = window.height as f32;

                    // Link virtual sphere rotation relative to window to make it feel nicer
                    let delta_x = rotation_move.x() / screen_width * std::f32::consts::PI * 2.0;
                    let delta_y = rotation_move.y() / screen_height * std::f32::consts::PI;

                    let delta_yaw = Quat::from_rotation_y(delta_x);
                    let delta_pitch = Quat::from_rotation_x(delta_y);

                    for (_, mut transform) in &mut camera_query.iter() {
                        let translation =
                            delta_yaw * delta_pitch * (transform.translation() - focus) + focus;
                        transform.set_translation(translation);

                        let look = Mat4::face_toward(transform.translation(), focus, Vec3::unit_y());
                        transform.set_rotation(look.to_scale_rotation_translation().1);
                    }
                }
                None => {
                    if let Some(pick) = pick_state.top(pick_group.0) { //  <------- use the pick to set the focus point of the camera rotation
                        state.focus = Some(*pick.position());
                    }
                }
            }

            return;
        }

        if state.focus.is_some() {
            state.focus = None;
        }
    }
}
```

Could you also kindly elaborate on the difference between `PickingMethod::Center` and `PickingMethod::ScreenSpace(Vec2::zero())`? is `Center` intended to allow casting a ray from something in world space and not the center of the screen? Apologies if I am misunderstanding the use cases here.

aevyrie:
Thanks for giving this a try!

PickingMethod::Center and PickingMethod::ScreenSpace(Vec2::zero()) are the same, but there is an important distinction between these variants. Because `Center` only needs to know the origin and direction of the source, the source can be anything with a `Transform`, but `ScreenSpace` requires a `Camera` because anything other than Vec2::zero() requires knowledge of the projection matrix.

> Multiple PickingSources have been added to pick group: 0

A pick group can only have one `PickingSource` at a time, this panic should only happen when a second picking source is added to a group. At a glance, I don't see that happening in your code, either there is a bug, or you area adding a second `PickingSource` to your group 0.

nickbryan:
Hey, my pleasure, always happy to do some testing 😄 

I have been over my code again and again and I cannot see how two sources would be getting added. Looking at your code, I can't see how there would be a bug with this either. It is really strange.

Thanks for the clarification. So does that mean it would be possible to have some component in the world with a transform the cast a ray from that? If so, how would it know which direction the ray would travel?

Sorry if I am getting over-involved and asking too many questions, I just find this very interesting haha.

nickbryan:
Ah, I have just run the debugger and it would seem it is multiple transforms that are causing the issue. Could it be that `mut transform_query: Query<&Transform>,` is causing the ray to be inserted multiple times for that group? Makes sense that it would be the same for `ScreenSpace` too as it is pulling in transforms too?

nickbryan:
Sorry, had a further thought on this and the ScreenSpace query is running twice because I have a ui camera and also a 3D camera in my game. This would cause it to iterate over both cameras with them both having a transform too. Is it just that `PickingSource` needs adding to them queries too?

aevyrie:
I don't think the transform running on both cameras is a problem, the only thing that really matters is 

```rust
.with(PickingSource::new(
        picking_group,
        PickingMethod::Center,
));
```

The picking plugin iterates over all entities with the PickingSource component, and generates a ray using this source for each group. If there are multiple PickingSources in the same group, it will cause the panic you see. Is it possible to share your repo so I can try running it locally?

aevyrie:
I think I found a bug, actually. It was constructing rays for both of your cameras, even though you only marked one of them as a PickingSource.

nickbryan:
Yeah sure thing https://github.com/nickbryan/zville. The camera code is [here](https://github.com/nickbryan/zville/blob/master/src/camera.rs#L30) and the mesh code [here](https://github.com/nickbryan/zville/blob/master/src/main.rs#L48).

My explinations were pretty poor before but I think the issue is with the following:
* In the `build_rays` function singature, the `mut camera_query: Query<(&Transform, &Camera)>` and `mut transform_query: Query<&Transform>,` are not constrained to a `PickingSource`.
* Every picking source `for pick_source in &mut pick_source_query.iter() {` will run for every entity with a transform and camera component `for (transform, camera) in &mut camera_query.iter() {` not just the ones that have a `PickingSource` attached. 
* `if pick_state.ray_map.insert(pick_source.group, pick_ray).is_some() {` will run for each of the entities insertign the group multiple times.

nickbryan:
Ah, haha. That was good timing!

nickbryan:
Yeah, just pulled the latest and that has worked!

nickbryan:
So with `PickingMethod::Center` does that mean it would be possible to have some component in the world with a transform then cast a ray from that? If so, how would it know which direction the ray would travel?

aevyrie:
> So with `PickingMethod::Center` does that mean it would be possible to have some component in the world with a transform then cast a ray from that? If so, how would it know which direction the ray would travel?

It's effectively identical to a camera transform, where the z axis is what you are looking down, and you can use the "look_at" function (I think).

nickbryan:
Ah of course! That's awesome dude, really nice job! Thanks for all the insights.

aevyrie:
Happy to help!

aevyrie:
@nickbryan Do you have suggestions on any better `PickingMethod` names? I didn't put too much thought into them, and it might be possible to improve clarity.

nickbryan:
Nah, I think the names you have make sense. The only one that might be ambiguous is Center but I can't think of an alternative.

Forgive me but I actually think there might still be a hidden issue with your fix. If I was to add another entity to either of these queries:
```rust
mut camera_query: Query<With<PickingSource,(&Transform, &Camera)>>,
mut transform_query: Query<With<PickingSource,&Transform>>,
```
then I think the issue will happen again. This is due to the loops running like "foreach pick state, loop over all of the components and insert a ray map for each component for this state" which will result in multiple entries for the same group.

I haven't tested this but I think it might need to be something like this instead:
```rust
fn build_rays(
    // Resources
    mut pick_state: ResMut<PickState>,
    cursor: Res<Events<CursorMoved>>,
    windows: Res<Windows>,
    // Queries
    mut camera_query: Query<(&PickingSource, &Transform, &Camera)>,
    mut transform_query: Query<(&PickingSource, &Transform)>,
) {
    // Collect and calculate pick_ray from all cameras
    pick_state.ray_map.clear();

    for (pick_source, transform, camera) in &mut camera_query.iter() {
        if let PickingGroup::Group(group_number) = pick_source {
            match pick_source.pick_method {
                PickingMethod::Cursor(window_id) => {
                    // Get the cursor position
                    let cursor_pos_screen: Vec2 = match pick_state.cursor_event_reader.latest(&cursor) {
                        Some(cursor_moved) => cursor_moved.position,
                        None => continue,
                    };

                    // Get current screen size
                    let window = windows.get(window_id).unwrap();
                    let screen_size = Vec2::from([window.width as f32, window.height as f32]);

                    // Normalized device coordinates (NDC) describes cursor position from (-1, -1, -1) to (1, 1, 1)
                    let cursor_pos_ndc: Vec3 =
                        ((cursor_pos_screen / screen_size) * 2.0 - Vec2::from([1.0, 1.0])).extend(1.0);

                    let camera_matrix = *transform.value();
                    let projection_matrix = camera.projection_matrix;
                    let (_, _, camera_position) = camera_matrix.to_scale_rotation_translation();

                    let ndc_to_world: Mat4 = camera_matrix * projection_matrix.inverse();
                    let cursor_position: Vec3 = ndc_to_world.transform_point3(cursor_pos_ndc);

                    let ray_direction = cursor_position - camera_position;

                    let pick_ray = Ray3D::new(camera_position, ray_direction);

                    if pick_state
                        .ray_map
                        .insert(pick_source.group, pick_ray)
                        .is_some()
                    {
                        panic!(
                            "Multiple PickingSources have been added to pick group: {}",
                            group_number
                        );
                    }
                }
                PickingMethod::ScreenSpace(coordinates_ndc) => {
                    let cursor_pos_ndc: Vec3 = coordinates_ndc.extend(1.0);
                    let camera_matrix = *transform.value();
                    let projection_matrix = camera.projection_matrix;
                    let (_, _, camera_position) = camera_matrix.to_scale_rotation_translation();

                    let ndc_to_world: Mat4 = camera_matrix * projection_matrix.inverse();
                    let cursor_position: Vec3 = ndc_to_world.transform_point3(cursor_pos_ndc);

                    let ray_direction = cursor_position - camera_position;

                    let pick_ray = Ray3D::new(camera_position, ray_direction);

                    if pick_state
                        .ray_map
                        .insert(pick_source.group, pick_ray)
                        .is_some()
                    {
                        panic!(
                            "Multiple PickingSources have been added to pick group: {}",
                            group_number
                        );
                    }
                },
                _ => (),
            }
        }

        for (pick_source, transform) in &mut transform_query.iter() {
            if let PickingGroup::Group(group_number) = pick_source {
                match pick_source.pick_method {
                    PickingMethod::Center => {
                        for transform in &mut transform_query.iter() {
                            let pick_position_ndc = Vec3::from([0.0, 0.0, 1.0]);
                            let source_transform = *transform.value();
                            let pick_position = source_transform.transform_point3(pick_position_ndc);

                            let (_, _, source_origin) = source_transform.to_scale_rotation_translation();
                            let ray_direction = pick_position - source_origin;

                            let pick_ray = Ray3D::new(source_origin, ray_direction);

                            if pick_state
                                .ray_map
                                .insert(pick_source.group, pick_ray)
                                .is_some()
                            {
                                panic!(
                                    "Multiple PickingSources have been added to pick group: {}",
                                    group_number
                                );
                            }
                        }
                    },
                    _ => (),
                }
            }
        }
    }
}
```
I think the reason it is currently working is that I only have one pick state registered with the system.

aevyrie:
As far as I'm aware, resources (like pick state) are intended to be global stores of state, so there should only ever be one instance of them.

I've structured the pick state with these assumptions and constraints in mind:

- There is only ever one pick state
- A `PickableMesh` can be assigned to many `PickingGroup`s
- There can only ever be one `PickingSource` in a `PickingGroup`
- The result of running the picking system is an ordered list of all intersections of each `PickingSource` with the `PickableMesh`s in its `PickingGroup`.

nickbryan:
Sorry I meant PickingSource not picking state. 

Currently, if I was to create a second camera and attach a second PickingSource to that camera with a second group then we would see the error message again on the first group. I think this is because on the first iteration of the picking_source_query the inner loop over the camera_query would actually loop over both cameras because the components match.

aevyrie:
Oh I see now, nice catch!

aevyrie:
I think I've fixed that up, and pre-empted some other bugs in the process, namely cursor events getting read (and removed from the queue) even if they aren't used by the picking source, e.g. when there are multiple windows open.

aevyrie:
Multi window is confirmed working
image

PR #46, opened 2020-09-24
Add intersection triangle to PickIntersection

guimcaballero:
This PR adds `world_triangle` and `mesh_triangle` to `PickIntersection`, to provide information about the triangle where the intersection happened.

I'm not sure if providing both the world and mesh triangle is needed, as it's just a transformation to get one from the other, but since both are already calculated I thought this was best to just add them both.

Edit: As requested in [Discord #help](https://discordapp.com/channels/691052431525675048/742884593551802431/758616863042437151)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -74,22 +74,14 @@ pub struct PickIntersection {
     entity: Entity,
     intersection: Ray3d,
     distance: f32,
-    mesh_triangle: Triangle,
     world_triangle: Triangle,
 }
 impl PickIntersection {
-    fn new(
-        entity: Entity,
-        intersection: Ray3d,
-        distance: f32,
-        mesh_triangle: Triangle,
-        world_triangle: Triangle,
-    ) -> Self {
+    fn new(entity: Entity, intersection: Ray3d, distance: f32, world_triangle: Triangle) -> Self {
         PickIntersection {
             entity,
             intersection,
             distance,
-            mesh_triangle,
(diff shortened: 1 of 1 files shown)


aevyrie:
Cool, thanks for this.

I'd prefer to store everything in world coordinates to stay consistent, and provide a function to transform into mesh coordinates as needed, or leave that part up to the user. My only concern is trying to keep structs lightweight as features are added, and compute as lazily as possible - e.g. only calculate things when the user actually needs it.

I haven't actually looked at your code yet, but I hope to soon. Do you mind making the changes I've mentioned?

guimcaballero:
Alright, makes sense. I've decided to not make a function to transform the world coords into mesh coords, as it would need to have the transform as a param, and then it's just using `transform_point3`. 

I've just left the world coordinates of triangle.

aevyrie (approved):
Looks great! Thank you for the changes.

aevyrie:
@guimcaballero Thanks for being patient, it took a bit longer than anticipated to get to this. 

FYI, there has also been some work to overhaul cursor events with #47. Cart (Mr. Bevy) was involved in this, and it looks like it will be the best way to handle reporting pick information going forward. I think this will mean moving all the pick data (intersection, triangle, etc) into components that get added to the picked entities instead of storing in a resource. If you have any thoughts on that, I'd appreciate your feedback in that thread so your voice is heard. Thank you!

PR #50, opened 2020-10-09
Update transform-based ray picking to use `GlobalTransform`

walterpie:
If an entity is parented it will show up in one place, but the ray
will hit it in a different place.  This commit fixes this issue.


aevyrie (approved):
Looks great, thank you!

Issue #47, opened 2020-09-26
Adding Cursor Events as Part of System


UnFlimFlammable:
I think a good improvement would be to add additional logic to handle 
on_click_down, on_click_up, on_hover, on_mouse_in, and on_mouse_out cursor events. 

I see two ways to accomplish this. One is to have ClickDown (and the other events) as components attached to each entity ala:

```
.with(ClickDown::new(|event, entity| {
    println!("I have been clicked most judiciously and with great conviction");
})
```
We could then create a system that would query all of the entities that possess ClickDown and run the attached closure as part of a system this library provides. I'm not sure this is the correct approach as you would be limited by the closure's scope. 

Alternatively we could provide a CursorEvents resource that contains all of the Cursor Event data such as hovering, meshes that have been click-down and click-upped etc. 

```
fn tic_tac_toe(mut commands: Commands, cursor_events: Res<CursorEvents>, mut comp_query: Query<&Tile>){
    //Logic here
    
}
```

The big draw of the second approach is matching the entities up to a click event in a performant manner. Currently you would have to do something like this:

```
    if let Some(pick_intersection) = pick_state.top(){
        //Handle Pick Data To Flip The Mesh
        for (tile, mut mesh) in tile_query.iter().iter() {
            
            if *tile == *tile_query2.get::<Tile>(pick_intersection.entity()).unwrap() {
                let cube = meshes.sphere.unwrap();
                *mesh = cube;
            }

        }
    }
```

At the least I think we should make that a utility method provided by the library, which would then make the resource-system oriented approach feasible and approachable to new developers.

```
    pick_state.find_matches(Query<Passed, To, The, Enclosing, System>) //This query must be a duplicate as internally they mutably borrow bevy resources. 
```

Would love to get someone's thoughts on this as I think it would be a big improvement to APi usability.

cart:
I think the `ClickDown` approach works for running logic on data captured in the closure / passed into it, but I have a feeling it won't be very useful in practice because you can't pass arbitrary ECS data into it.

CursorEvents would work for "generic" use cases, but as you mentioned it would have performance implications because you need to do a full scan through events you might not care about. This probably won't be a huge deal, but its still worth thinking about.

In general this is very similar to the "click event" problem we solved in Bevy UI. Currently we're using an Interaction component on each "interactable" ui element, and we set the state on it (hovered, clicked, none) whenever something changes.

The "change tracking" queries then allow us to write systems that react to changes in state:
[...]

PR #52, opened 2020-10-13
Cursor Events

aevyrie:
Working on implementing cursor events and moving pick state information from a resource to entities. See #47.


src/highlight.rs
```diff
@@ -37,89 +37,83 @@ impl Default for PickHighlightParams {
...
-    pick_state: Res<PickState>,
     mut materials: ResMut<Assets<StandardMaterial>>,
     highlight_params: Res<PickHighlightParams>,
     // Queries
-    mut query_picked: Query<(
-        &mut HighlightablePickMesh,
-        &PickableMesh,
-        &Handle<StandardMaterial>,
-        Entity,
-    )>,
```
nicbarker, line 69: Ah yes I see here, you don't get the conflicting query cause you've rolled it into a single one 🙂

aevyrie: I wish I had noticed and saved you the time, I've been working on this sporadically and it's hard to keep track of what has diverged from master. 😅

nicbarker: Haha no worries, looking at this mod has been really useful in getting a better idea for how bevy works / what it doesn't like 😂

PR #58, opened 2020-11-01
Update readme to reflect PickingGroup -> PickGroup and standardize naming

nicbarker:
Hello 👋  another one!

It seems that `PickingGroup` has been renamed to `PickGroup` in the code. I've updated the README for consistency.
I also did a small rename of a `picking_group` argument in the code to `pick_group` for consistency.

PR #59, opened 2020-11-02
Update for compatibility with Bevy 0.2

nicbarker:
This PR includes a number of small changes to realign with the new bevy master, including:
- Update use of `VertexAttribute` to the new `Map` structure
- Simplify query iterators as they've been converted to real iterators (we don't need `to_iter().to_iter()` anymore)
- Migrate usages of `query.entity` and `query.component` to `query.get` and `query.get_component`
- Split pick_highlighting into two systems to avoid conflicting queries:
Screen Shot 2020-11-02 at 4 36 52 PM

### DIFF
--- a/examples/interactable_cube.rs
+++ b/examples/interactable_cube.rs
@@ -64,8 +64,8 @@ fn setup(
         });
 }
 
-fn interactable_demo(mut imesh_entities: Query<&InteractableMesh>) {
-    for imesh in imesh_entities.iter().iter() {
+fn interactable_demo(imesh_entities: Query<&InteractableMesh>) {
+    for imesh in imesh_entities.iter() {
         if imesh.mouse_hover {
             //println!("Hovering!");
         }
@@ -78,19 +78,16 @@ fn interactable_demo(mut imesh_entities: Query<&InteractableMesh>) {
             println!("Mouse Exited");
         }
 
-        match imesh.mouse_down(MouseButton::Left) {
-            Some(v) => println!("Left Mouse Button is Down"),
-            None => (),
+        if imesh.mouse_down(MouseButton::Left).is_some() {
+            println!("Left Mouse Button is Down");
         }
 
-        match imesh.mouse_just_pressed(MouseButton::Left) {
-            Some(v) => println!("Left Mouse just Clicked"),
-            None => (),
+        if imesh.mouse_just_pressed(MouseButton::Left).is_some() {
+            println!("Left Mouse just Clicked");
         }
 
-        match imesh.mouse_just_released(MouseButton::Left) {
-            Some(v) => println!("Left Mouse just Released"),
-            None => (),
+        if imesh.mouse_just_released(MouseButton::Left).is_some() {
+            println!("Left Mouse just Released");
         }
     }
 }
--- a/src/debug.rs
+++ b/src/debug.rs
@@ -46,15 +46,15 @@ fn update_debug_cursor_position(
     // Set the cursor translation to the top pick's world coordinates
     for (_group, top_pick) in pick_state.top_all() {
         let transform_new = top_pick.intersection.to_transform();
-        for mut transform in &mut query.iter() {
+        for mut transform in &mut query.iter_mut() {
             *transform = Transform::from_matrix(transform_new);
         }
-        for mut draw in &mut visibility_query.iter() {
+        for mut draw in &mut visibility_query.iter_mut() {
             draw.is_visible = true;
         }
     }
     if pick_state.top_all().is_empty() {
-        for mut draw in &mut visibility_query.iter() {
+        for mut draw in &mut visibility_query.iter_mut() {
             draw.is_visible = false;
         }
     }
--- a/src/highlight.rs
+++ b/src/highlight.rs
@@ -72,10 +72,10 @@ pub fn pick_highlighting(
         &SelectablePickMesh,
         &Handle<StandardMaterial>,
     )>,
-    mut query_selectables: Query<&SelectablePickMesh>,
+    query_selectables: Query<&SelectablePickMesh>,
 ) {
     // Query selectable entities that have changed
-    for (mut highlightable, selectable, material_handle) in &mut query_selected.iter() {
+    for (mut highlightable, selectable, material_handle) in &mut query_selected.iter_mut() {
         let current_color = &mut materials.get_mut(material_handle).unwrap().albedo;
         let initial_color = match highlightable.initial_color {
             None => {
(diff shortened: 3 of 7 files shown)


examples/interactable_cube.rs
```diff
@@ -78,19 +78,16 @@ fn interactable_demo(mut imesh_entities: Query<&InteractableMesh>) {
...
-            None => (),
+        if imesh.mouse_just_pressed(MouseButton::Left).is_some() {
+            println!("Left Mouse just Clicked");
         }
 
-        match imesh.mouse_just_released(MouseButton::Left) {
-            Some(v) => println!("Left Mouse just Released"),
-            None => (),
+        if imesh.mouse_just_released(MouseButton::Left).is_some() {
+            println!("Left Mouse just Released");
```
nicbarker, line 90: Clippy was complaining about a single case `match` as well as `v` being unused and recommended that it was switched to if statement, let me know if this is ok, I can always revert it

src/debug.rs
```diff
@@ -46,15 +46,15 @@ fn update_debug_cursor_position(
     // Set the cursor translation to the top pick's world coordinates
     for (_group, top_pick) in pick_state.top_all() {
         let transform_new = top_pick.intersection.to_transform();
-        for mut transform in &mut query.iter() {
+        for mut transform in &mut query.iter_mut() {
```
nicbarker, line 49: There is now a distinction between `iter` and `iter_mut`

src/highlight.rs
```diff
@@ -57,25 +57,17 @@ impl Default for HighlightablePickMesh {
...
-    pick_state: Res<PickState>,
     mut materials: ResMut<Assets<StandardMaterial>>,
     highlight_params: Res<PickHighlightParams>,
     // Queries
-    mut query_picked: Query<(
-        &mut HighlightablePickMesh,
-        &PickableMesh,
-        &Handle<StandardMaterial>,
-        Entity,
-    )>,
```
nicbarker, line 69: There are no real changes here, this is simply splitting into two systems to avoid bevy crashing at runtime with `pick_highlighting has conflicting queries`. I believe it's because there are two queries (`query_picked` and `query_selected`) that are both trying to access `HighlightablePickMesh` as `mut`, but I could be wrong.

src/interactable.rs
```diff
@@ -92,8 +92,7 @@ pub fn cursor_events(
 ) {
     //Go through the pick state and find the
     q_imesh
-        .iter()
-        .iter()
```
nicbarker, line 96: We don't need double `.iter` because of cart's changes 👍

src/lib.rs
```diff
@@ -16,7 +16,7 @@ use bevy::{
     prelude::*,
     render::{
         camera::Camera,
-        mesh::{Indices, VertexAttribute, VertexAttributeValues},
```
nicbarker, line 19: `VertexAttribute` no longer exists, it's been replaced with a Map and some const strings on `Mesh`

src/lib.rs
```diff
@@ -383,8 +384,8 @@ fn pick_mesh(
             let vertex_positions: Vec<[f32; 3]> = mesh
                 .attributes
                 .iter()
-                .filter(|attribute| attribute.name == VertexAttribute::POSITION)
```
nicbarker, line 386: attribute.name has been replaced with the first item in a tuple of `(str, VertexAttributeValues)`

src/raycast.rs
```diff
@@ -103,8 +103,8 @@ impl From<&Mesh> for BoundingSphere {
         }
         let mut vertex_positions = Vec::new();
         for attribute in mesh.attributes.iter() {
-            if attribute.name == VertexAttribute::POSITION {
-                vertex_positions = match &attribute.values {
```
nicbarker, line 107: Same as above

src/select.rs
```diff
@@ -33,12 +33,12 @@ pub fn select_mesh(
...
     if mouse_button_inputs.just_pressed(MouseButton::Left) {
         // Deselect everything
-        for mut selectable in &mut query.iter() {
+        for mut selectable in &mut query.iter_mut() {
             selectable.selected = false;
         }
 
         for (_group, pick) in pick_state.top_all() {
-            if let Ok(mut top_mesh) = query.get_mut::<SelectablePickMesh>(pick.entity) {
+            if let Ok(mut top_mesh) = query.get_mut(pick.entity) {
```
nicbarker, line 41: `get_mut` now has nice ergonomic typing so we don't need to specify

aevyrie (commented):
Thanks for doing this! My only concern lies with the split highlight system. Have you tested in 3d_scene to make sure highlighting still takes priority over selection?

src/lib.rs
```diff
@@ -33,7 +33,8 @@ impl Plugin for PickingPlugin {
             .add_system(pick_mesh.system())
             .add_system(cursor_events.system())
             .add_system(select_mesh.system())
-            .add_system(pick_highlighting.system());
+            .add_system(pick_highlighting.system())
+            .add_system(pick_selecting.system());
```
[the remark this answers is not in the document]
nicbarker: Ah yes of course, nice that makes sense. I'll double check the 3d test scene and update if necessary

nicbarker:
Hey @aevyrie could you take another look? I came up with an alternative idea - instead of splitting the system, I removed the query of `HighlightablePickMesh` from `query_selected`. As a result, the color of the mesh will only be reset to the original color when iterating through `query_picked` (hovered) rather than through `query_selected`.

I think this achieves the same outcome as previous, right?

aevyrie:
> As a result, the color of the mesh will only be reset to the original color when iterating through `query_picked` (hovered) rather than through `query_selected`.

Doesn't this create a problem when clicking to select or deselect? The correct behavior would be that on click, regardless of what is currently hovered, a previously selected item should deselect and go to original color. As you've described, clicking on various meshes would not cause the appearance to update unless the newly deselected mesh is hovered over.

aevyrie:
(I swear I'm not trying to be difficult! 😄)

For what it's worth, I have a significant plugin refactor in the works in #52, which is working on current branch without issues, and is wrapping up quicker than anticipated. I'm almost tempted to suggest you not bother with trying to make these fixes if I'm going to come through and write over it all due to the inevitable huge merge conflicts. Maybe it's more worth your time to hold off, then merge in the renaming changes from #58 once I have #52 in?

Let me know what you think. I apologize if I've wasted your time here.

nicbarker:
> (I swear I'm not trying to be difficult! 😄)

No not at all 🙂  If it's easier for you to do the updates yourself, that's totally fine!

nicbarker:
closing in favour of https://github.com/aevyrie/bevy_mod_picking/pull/52

Issue #67, opened 2020-11-26
Mouse picking intersection doesn't update on camera move


Sheepyhead:
Hello, lovely plugin you have here! I was just implementing some simple movement in my game using this, and I ran into a minor issue. See I have a cube that has to move along a plane when a part of the plane's surface is clicked, and it's supposed to move to the point of the plane. The issue I'm facing is that the camera moves with it, and the intersection does not get updated unless I move the mouse, despite the fact that the mouse obviously intersects with the mesh in a different spot due to the new camera position. See attached gif for an example. 

I'm using the InteractableMesh and everything else works swimmingly, I figure this might be an oversight in how the InteractableMesh::mouse_down_event interacts with PickableMesh::intersection. 

I'm using version 0.2.0 so if this has already been fixed I apologize

![wMFEzTgViq (Custom) (2)](https://user-images.githubusercontent.com/15895891/100344720-8cb8cb00-2fe1-11eb-9dda-68e201f8a5a2.gif)

aevyrie:
Hey, thanks for the report! I'll have to see if this is happening in the debug cursor, or in the actual picking systems. I think the intent here is to only compute picks during a mouse movement, instead of running constantly. Some ways we could specify the always-on behavior:

1. In your world movement systems, send a callback to the picking system, letting it know it will need to run a check this frame because you've moved things around.
2. When setting up the `PickingSource`, there could be an "always-on" mode you specify.

Let me know if you have any thoughts on this! 🙂

Sheepyhead:
> 
> 
> Hey, thanks for the report! I'll have to see if this is happening in the debug cursor, or in the actual picking systems. I think the intent here is to only compute picks during a mouse movement, instead of running constantly. Some ways we could specify the always-on behavior:
> 
>     1. In your world movement systems, send a callback to the picking system, letting it know it will need to run a check this frame because you've moved things around.
> 
>     2. When setting up the `PickingSource`, there could be an "always-on" mode you specify.
> 
> 
> Let me know if you have any thoughts on this! 🙂

As shown in the example it also applies to the intersection returned by InteractableMesh, so the debug cursor is accurately showing the intersection it's returning. Now what I'd personally find most useful is to have a setting for PickingSource that it keeps updating the intersection even when a mouse button is held, personally I'd consider that the expected behavior, but leaving it as an option is fine

aevyrie:
[...]

Issue #64, opened 2020-11-09
Mouse picking is not updated when mouse is not moved and object is deleted


Nickan:
Is there a way not to depend on the cursor_events.latest() when updating the picking on line 263 on lib.rs?

I am thinking of adding a field in PickState the previous_mouse_pos and use it when cursor is not moved, but I would like to know if that's the possible solution, thanks.

Sample code is here: https://github.com/Nickan/bevytest/blob/ad375b83351cd6d698c316ff7d00e82db3f7a824/src/bevy_mod_picking_test.rs

aevyrie:
Hi @Nickan, I'm not sure I follow. Is this a bug you found (title)?

Nickan:
I don't really think it is a bug, but a feature request, sorry for being vague, I will MR later what I want to happen. 

Basically, I want to get the top pick, regardless if the cursor is moving or not.

aevyrie:
Yeah, if you can demonstrate the current behavior and the behavior you are looking for, that will really help me understand. Thanks!

aevyrie:
Closing with merge of #68

PR #68, opened 2020-12-05
Added option to always update the mouse picking via saving the last m…

Nickan:
Hi! We implemented the ff:
- Added option to always update the mouse picking even when mouse is not moved via saving the last mouse position
- By default it is false.
- Snippet to set as always_on:
`.with(PickSource {
      always_on: true,
      ..Default::default()
    })`
- Solves issues: https://github.com/aevyrie/bevy_mod_picking/issues/64 and https://github.com/aevyrie/bevy_mod_picking/issues/67

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -274,12 +274,12 @@ fn build_rays(
                         }
                     }
                     None => {
-                      if (pick_source.always_on) {
-                        pick_state.last_cursor_pos
-                      } else {
-                        continue
-                      }
-                    },
+                        if (pick_source.always_on) {
+                            pick_state.last_cursor_pos
+                        } else {
+                            continue;
+                        }
+                    }
                 };
                 pick_state.last_cursor_pos = cursor_pos_screen;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -277,16 +277,12 @@ fn build_rays(
                             continue;
                         }
                     }
-                    None => {
-                        match update_picks {
-                            UpdatePicks::Always => {
-                                pick_state.last_cursor_pos
-                            }
-                            UpdatePicks::OnMouseEvent => {
-                                continue;
-                            }
+                    None => match update_picks {
+                        UpdatePicks::Always => pick_state.last_cursor_pos,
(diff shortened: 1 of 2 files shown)


aevyrie:
Thanks! What is the impact of defaulting to always on? 

For example, if the window is not active, will it continue to check for intersections every frame? I think the answer is yes.

I ask because based on @Sheepyhead's use case, they would need it to be enabled, and it might actually make sense if this is default behavior. However, my only worry is that it will always be running even when not needed. This isn't an issue now, because mouse events are not sent when the window is inactive.

src/lib.rs
```diff
@@ -186,16 +188,18 @@ pub enum PickMethod {
 
 /// Marks an entity to be used for picking
 pub struct PickSource {
-    groups: Option<Vec<Group>>,
-    pick_method: PickMethod,
-    cursor_events: EventReader<CursorMoved>,
+    pub groups: Option<Vec<Group>>,
+    pub pick_method: PickMethod,
+    pub always_on: bool,
```
aevyrie, line 193: The `always_on` variable is only used for the `PickMethod::CameraCursor`, could you move it into that variant definition?

In addition To be a bit more self documenting, I would prefer an enum instead of a bool:
```rust
enum update_picks {
    always,
    on_mouse_event,
}
```

aevyrie (changes requested):

src/lib.rs
Nickan: Done, yes you are right.

I did not make it a default behavior though, will make it if you are certain we make it so. Thanks.

Nickan: The CI fails because the master branch is pointed to the bevy master, how should we deal with this?

aevyrie: Looks like PR #69 will fix the CI issue. I'll merge that and we can re-run the tests.

aevyrie:
Looks like the multiple_window example needs a small fix with this change.
https://github.com/aevyrie/bevy_mod_picking/pull/68/checks?check_run_id=1540877088#step:6:405

aevyrie (approved):

PR #70, opened 2020-12-14
disable bevy default features

mockersf:
this speeds up compile time, and allow usage of this library in a wasm build

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,5 +12,5 @@ categories = ["game-engines", "rendering"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-# bevy = "^0.3"
-bevy = { git = "https://github.com/bevyengine/bevy", branch = "master" }
+# bevy = { version = "^0.3", default-features = false, features = ["render"] }
+bevy = { git = "https://github.com/bevyengine/bevy", branch = "master", default-features = false, features = ["render"] }
--- a/examples/multiple_windows.rs
+++ b/examples/multiple_windows.rs
@@ -39,8 +39,8 @@ fn setup(
     create_window_events.send(CreateWindow {
         id: window_id,
         descriptor: WindowDescriptor {
-            width: 800,
-            height: 600,
+            width: 800.0,
+            height: 600.0,
             vsync: false,
             title: "second window".to_string(),
             ..Default::default()


aevyrie:
Looks like a small change in master caused a test failure in `multiple_windows` on lines 42/43 (add a ".0" to make window size a float).

Does this change have any other implications you're aware of? I'm not well versed in the use of cargo features. I assume all this is doing is minimizing the number of feature dependencies, which is useful in applications like wasm.

mockersf:
It doesn't change anything except reducing the number of dependencies for applications that are already limiting the number of dependencies

The important (and undocumented) thing about rust features is that cargo treats them as additive, so it will use the union of all features enabled to a dependency. you can't have mutually exclusive features, you can't have features enabled for one transitive dependency and not for the other (ie if I depend on bevy and on bevy_mod_picking which depends on bevy, bevy will be compiled with the union of enabled features from both dependencies).

I tested this PR in a wasm build and it works 👍

aevyrie:
Thanks for the explanation - really appreciate that! Would you mind either:

-  including the small fix to resolve the failed test
-  or, pointing to a previous bevy git commit in the cargo.toml, prior to this breaking change

I can then merge in this PR. Thanks!

mockersf:
test fixed!

aevyrie:
Turns out this is causing a runtime panic for the examples: 
```shell
thread 'thread 'mainthread '' panicked at 'Compute Task Pool (1)Resource does not exist alloc::boxed::Box<dyn bevy_render::renderer::render_resource_context::RenderResourceContext>.' panicked at 'Compute Task Pool (3)', Resource does not exist alloc::boxed::Box<dyn bevy_render::renderer::render_resource_context::RenderResourceContext>.C:\Users\aevyr\.cargo\git\checkouts\bevy-f7ffde730c324c74\45e2be3\crates\bevy_ecs\src\resource\resources.rs', ' panicked at 'C:\Users\aevyr\.cargo\git\checkouts\bevy-f7ffde730c324c74\45e2be3\crates\bevy_ecs\src\resource\resources.rs::Resource does not exist alloc::boxed::Box<dyn bevy_render::renderer::render_resource_context::RenderResourceContext>.282282::32',
32C:\Users\aevyr\.cargo\git\checkouts\bevy-f7ffde730c324c74\45e2be3\crates\bevy_ecs\src\resource\resources.rsnote: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
:
282:32
thread 'main' panicked at 'Task thread panicked while executing.: Any', C:\Users\aevyr\.cargo\git\checkouts\bevy-f7ffde730c324c74\45e2be3\crates\bevy_tasks\src\task_pool.rs:75:18
```

I think this means we need to figure out which dependency is needed, then add it as a required feature for examples.

mockersf:
sorry about that, it's done in #71

PR #71, opened 2020-12-15
add features for examples after disabling bevy default features

mockersf:
forgot to add needed features for examples: https://github.com/aevyrie/bevy_mod_picking/pull/70#issuecomment-745027081

example `multiple_windows` is still panicking, but then it also does on bevy: https://github.com/bevyengine/bevy/issues/1069


aevyrie:
Nice, was about to create a PR with nearly identical changes. I've had to learn a bit more about cargo features. 😄 
Thanks for pointing to the bevy issue, I was just starting to look into the panic.

PR #80, opened 2020-12-17
Bounding sphere

aevyrie:
Bounding sphere generation and culling optimization, #2


aevyrie:
Initial tests of the stress test case show an 11X improvement in performance. This is extremely sensitive to the test scenario.

PR #87, opened 2021-01-06
 Using log::trace instead of println

halzy:
Bevy already includes a reference to the log crate, so anyone using bevy would already have it set up.

Switching from println() to log::trace() helps those that are not using the bevy logger and are filtering log outputs.

### DIFF
--- a/src/debug.rs
+++ b/src/debug.rs
@@ -12,7 +12,7 @@ impl Plugin for DebugPickingPlugin {
 
 fn get_picks(pick_state: Res<PickState>) {
     if cfg!(debug_assertions) {
-        println!("Top entities:\n{:#?}", pick_state.top_all())
+        trace!("Top entities:\n{:#?}", pick_state.top_all())
     }
 }
 


aevyrie:
This is a great suggestion, thanks.
Instead of using an external dependency, can we make use of Bevy's `LogDiagnosticsPlugin`?

aevyrie:
Hey @halzy, any thoughts on my comment?

halzy:
Hello, apologies for the delay. `LogDiagnosticsPlugin` doesn't seem to expose the `log::*` functions which are needed.

aevyrie:
I asked around in the Bevy discord, and it turns out `trace!` is available as part of the bevy prelude import. You can remove the change to cargo.toml, and change `log::trace!()` to just `trace!()`.

aevyrie:
@halzy it looks like bevy/master is now using these too: https://github.com/bevyengine/bevy/blob/master/examples/app/logs.rs

```rust
// here is how you write new logs at each "log level" (in "least important" to "most important" order)
trace!("very noisy");
debug!("helpful for debugging");
info!("helpful information that is worth printing by default");
warn!("something bad happened that isn't a failure, but thats worth calling out");
error!("something failed");
```

I would strongly prefer to use this method to mirror bevy best practices. Does this seem to fit your needs? This repo has changed significantly over the last week or two, I am happy to make these changes myself if you don't have the time.

halzy:
I'm going to go ahead and close the PR. The repo seems to have changed quite a bit and I've been heads down at work. Apologies for not seeing these comments sooner. Thank you for all you've done!

aevyrie:
No worries! Thanks for bringing this to my attention. 🙂

PR #92, opened 2021-01-19
enable x11 feature for example_deps

jakobhellermann:
Without it, the example would fail to compile on linux.


aevyrie:
Great, thanks!

PR #111, opened 2021-12-30
Expose Intersection type from bevy_mod_raycast

alexlegg:
This is the type returned by intersect_top() on PickingCamera.


alexlegg:
Oops, just noticed everything is exposed by functions. No need for this.

PR #125, opened 2022-04-16
Updated to bevy 0.7

obsgolem:
The biggest change affecting the code is the need to replace references to the camera's window to references to the RenderTarget.

Tested using cargo's `patch` feature using https://github.com/aevyrie/bevy_mod_raycast/pull/43.

On my machine stress_test doesn't appear to render anything. I don't know if my changes are responsible, but disabling all the non-default plugins left the scene empty.

* [ ] Merge after https://github.com/aevyrie/bevy_mod_raycast/pull/43 and after 0.4 release is done.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bevy_mod_picking"
-version = "0.5.4"
+version = "0.6.0"
 authors = ["Aevyrie <aevyrie@gmail.com>"]
 edition = "2021"
 license = "MIT"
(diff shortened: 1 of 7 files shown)


aevyrie:
Thanks @obsgolem! I'll take a look at merging these early next week.

obsgolem:
Note that I didn't do the perhaps more obvious thing of merging main on these PRs, mainly because I didn't think of it. I think my solution to the `window` vs `target` issue is slightly more robust than the one on main. Might be a good idea to port those changes over to main and just make this a merge of main instead.

Issue #108, opened 2021-11-15
Selecting family as one object


pop:
I'd love to be able to group multiple Meshes into one Pickable entity. i.e., if I spawn a Parent entity with multiple children meshes, selecting any one of the children selects the entire hierarchy (parent and siblings).

My use-case is that I am importing a GLTF Scene which has Nodes which are made of multiple Meshes. I want to treat each GLTF Node as one Pickable entity. I've tried implementing this locally but haven't made progress.

If you (A) know of a workaround or (B) know where in the code I could add this feature I would gladly make this a pull request.

I expect this would be implemented with the following psuedocode:

1. When a mesh is selected, check if the mesh has a parent.
2. Update the state for all of that parent's children to match the selected child's state.
3. Hierarchy's could be implemented with a `PickableHierarchyBundle` on the parent and/or children -- or just tag the parent (as well as the children) with the `PickableBundle`.

I'd love to discuss design if I'm [not] on the right track.

P.S., I love the plugin! I'm using it in a game and it's made a point-and-click mechanic very easy!

pop:
Update, I got a very rough and dirty version of this feature implemented. I'd love feedback on the idea and the implementation.

[Branch Here](https://github.com/aevyrie/bevy_mod_picking/compare/master...pop:pop/pickable-hierarchy)

aevyrie:
Thanks @pop, sorry for the delayed response. This looks really useful! If you're still willing to implement this, I could help out on a PR branch.

pop:
I just picked up my gamedev project again, I should have a PR up for this in a few days.

aevyrie:
Cool! Looking forward to seeing what you come up with. 🙂

sildude:
Would love to be able to simply spawn a gltf scene and have this work out of the box.

ericvrp:
Is there any news about this?

pop:
@ericvrp I struggled with the implementation details but I finally got a _rough_ implementation done in #126. Check it out if you're interested, I'd love feedback on the API choices and implementation.

pop:
@sildude The implementation in #126 doesn't work with selecting GLTFs, but after digging into the code I see how one _would_ add GLTF selections.

The interface would require tagging a parent entity with `PickableMesh` or a new `PickableGltf` and spawning your GLTF mesh/scene as a child of that entity. Something like this:

```rust
let scene = assets.load("whatever.gltf#Scene0");

// cube
commands.spawn_bundle((
[...]

PR #126, opened 2022-04-17
Family Selection

pop:
# Overview
Implements _rough_ family selection

The goal of this feature is to be able to treat a family of entities as one "pickable object".

Uses for this would be spawning a scene of objects that are disjoint meshes but you intend to treat as one "thing".

Note two API choices being made:
1. All entities in a pickable faimliy must have the PickableMesh component.
2. There is no explicit PickableFmaily component, meaning a family of PickableMesh entities will be selected as one entity.

Currently the functionality is gated with a feature flag, but I'm considering moving to a Bundle/Plugin option.
Then you can choose on a per-entity basis if it's in a "Family Selection" group instead of it being global.

# Examples

```
cargo run --example family --features family
```

# Tasks

- [ ] Resolve conflicts with `main`. This branch was based on `v0.5.4`.
- [x] Add feature to enable/disable this functionality.
- [ ] Finalize API
- [ ] Finalize implementation
- ~GLTF scene example.~

After logging into GLTF loading I think that should be a different body of work.

Refs: #108

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bevy_mod_picking"
-version = "0.6.0"
+version = "0.6.1"
 authors = ["Aevyrie <aevyrie@gmail.com>"]
 edition = "2021"
 license = "MIT"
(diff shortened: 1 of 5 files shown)


aevyrie:
Closing as stale. I believe many of the changes in the beta branch solve the problems addressed here.

Issue #124, opened 2022-04-13
PausedForBlockers get stuck in true state, if UI is removed while it is blocked


BessBrahmin:
Hello, I ran to an issue with the picker not working on my own project. I tracked the issue into the method "pause_for_picking_blockers" setting pause.0 = true during the game being in the main menu. But when I went to game from the menu the pause would be stuck on true, since the for statement in the method. That should make it false never triggers due there being no UI elements outside of the Main menu.

I fixed this myself by moving the pause.0 = false to the end of the method and outside of the for statement, since the pause.0 = true calls return right after it anyway.

Since I'm not familiar with this package I don't want to make a pull request, since my fix might actually cause many issues I couldn't predict. So I made this ticket instead, maybe someone who is familiar with the code is able to make a proper fix for this issue.

aevyrie:
@BessBrahmin please go ahead and make a PR, it will be easier to review if it has any unintended side effects. Thank you!

docDock:
This happens to me as well. I will submit a pr that fixes this.

PR #132, opened 2022-05-22
FIx for issue #124 submitted by BessBrahmin.

docDock:
After despawning all ui elements PauseForBlockers resource would become permanently set to true. This fixes issue by checking if there stting PauseForBlockers to false if there are no ui interactions.

### DIFF
--- a/src/focus.rs
+++ b/src/focus.rs
@@ -40,7 +40,9 @@ pub fn pause_for_picking_blockers(
         Query<&Interaction, Or<(With<Node>, With<PickingBlocker>)>>,
     )>,
 ) {
+    let mut no_ui_interactions = true;
     for ui_interaction in interactions.p1().iter() {
+        no_ui_interactions = false;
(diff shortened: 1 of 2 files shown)


aevyrie (changes requested):
I think this whole function can be simplified to:
```rs
paused.0 = false;
for ui_interaction in interactions.p1().iter() {
    if *ui_interaction != Interaction::None {
        for (mut interaction, hover, _, _) in &mut interactions.p0().iter_mut() {
            if *interaction != Interaction::None {
                *interaction = Interaction::None;
            }
            if let Some(mut hover) = hover {
                if hover.hovered {
                    hover.hovered = false;
                }
            }
        }
        paused.0 = true;
        return;
    }
}
```

docDock:
Updated

aevyrie (approved):

aevyrie:
Thanks!

PR #137, opened 2022-06-01
Gigantic Refactor

aevyrie:
Fully decouples picking from both input and picking backend. Adds the ability to place event listeners on entities, and easy event-forwarding to entities on interaction.


crates/bevy_picking_core/src/focus.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub fn hovered(&self) -> bool {
+        self.hovered
+    }
+}
+
+/// Marker component for entities that, whenever their [Interaction] component is anything other
+/// than `None`, will suspend highlighting and selecting [PickableTarget]s. Bevy UI [Node]s have this
+/// behavior by default.
+#[derive(Component)]
+pub struct PickingBlocker;
```
alice-i-cecile, line 24: This seems extremely nice; I'd like this upstreamed.

src/lib.rs
```diff
@@ -1,251 +1,252 @@
...
-                        mesh_events_system
-                            .label(PickingSystem::Events)
-                            .after(PickingSystem::Selection),
-                    ),
-            );
-    }
+/// Makes an entity pickable.
+#[derive(Bundle, Default)]
+pub struct PickableBundle {
+    /// The entity's configurable [`PickLayer`](focus::PickLayer)
```
alice-i-cecile, line 158: These docs will be more useful if you provide a couple words expanding on what the component does.

src/lib.rs
```diff
@@ -1,251 +1,252 @@
...
+    /// Tracks entity [`PickHighlight`](highlight::PickHighlight) state.
+    pub highlight: highlight::PickHighlight,
 }
 
-pub struct HighlightablePickingPlugins;
-impl PluginGroup for HighlightablePickingPlugins {
-    fn build(&mut self, group: &mut PluginGroupBuilder) {
-        group.add(CustomHighlightPlugin(StandardMaterialHighlight));
-        group.add(CustomHighlightPlugin(ColorMaterialHighlight));
+/// Components needed to build a pointer
```
alice-i-cecile, line 172: This should mention that multiple pointers can be active at once, and that each one is an entity.

src/lib.rs
```diff
@@ -1,251 +1,252 @@
...
+    /// Tracks the pointer's button press state.
+    pub click: pointer::PointerPress,
+    /// Tracks the pointer's interaction state.
+    pub interaction: output::PointerInteraction,
+    #[cfg(feature = "selection")]
+    /// Tracks whether the pointer's multiselect is active.
+    pub multi_select: selection::PointerMultiselect,
+}
+impl PointerBundle {
+    /// Create a new pointer with the provided `id`.
```
alice-i-cecile, line 188: It's not clear here how these should be generated, or how users can ensure they're unique.

Should you be using an atomic here instead of letting users pass in their own ID?

aevyrie: `PointerID` is a type, see the documentation on the type. Users only need to use this if they don't use the provided defaults (mouse + touches), or they want to add a software pointer.

examples/bubbling.rs
```diff
@@ -0,0 +1,90 @@
...
+        .add_plugin(backends::RaycastPlugin)
+        .add_plugin(DebugEventsPlugin)
+        .add_startup_system(setup)
+        .add_system(handle_events)
+        .run();
+}
+
+struct DeleteMe(Entity);
+impl EventFrom for DeleteMe {
+    fn new(event_data: &mut EventData<impl IsPointerEvent>) -> Self {
```
alice-i-cecile, line 21: These examples will be much clearer if you avoid using `Self` in these impl blocks.

examples/bubbling.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+struct GreetMe(Entity);
+impl EventFrom for GreetMe {
+    fn new(event_data: &mut EventData<impl IsPointerEvent>) -> Self {
+        Self(event_data.target())
+    }
+}
+
+fn handle_events(
```
alice-i-cecile, line 36: This should be two systems.

examples/bubbling.rs
```diff
@@ -0,0 +1,90 @@
+use bevy::prelude::*;
```
alice-i-cecile, line 5: This example needs some long-form comments (maybe at the top) explaining exactly what's going on to avoid magic.

crates/backends/bevy_picking_shader/NOTES.md
```diff
@@ -0,0 +1,20 @@
...
+# Impl
+
+group all cameras by `render_target`, note that pointers also correspond to render target
+sort cameras by `priority`
+create a picking camera as a child of the camera, so it follows it. 
+set the target of the picking camera to be a texture
+
+Is the above just duplicating work already done in bevy_render?
+
+Instead, need to add a node to the render graph that takes the output from the vertex shader, and runs th picking frag shader, writing to a texture.
```
colepoirier, line 20:
```suggestion
Instead, need to add a node to the render graph that takes the output from the vertex shader, and runs the picking frag shader, writing to a texture.
```

crates/bevy_picking_core/src/focus.rs
```diff
@@ -0,0 +1,141 @@
...
+/// interaction state, in this case, not just whether the pointer happens to be over the entity.
+#[derive(Debug, Deref, DerefMut, Default)]
+pub struct HoverMap(pub HashMap<PointerId, HashSet<Entity>>);
+
+/// Coalesces all data from inputs and backends to generate a map of the currently hovered entities.
+pub fn update_focus(
+    // Inputs
+    focus: Query<&FocusPolicy>,
+    pick_layers: Query<&PickLayer>,
+    pointers: Query<(&PointerId, &PointerInteraction)>, // <- what happened last frame
```
colepoirier, line 56: This function would benefit from more extensive docs.

crates/bevy_picking_core/src/output.rs
```diff
@@ -0,0 +1,628 @@
...
+}
+impl DerefMut for PointerInteraction {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.map
+    }
+}
+
+/// Can be implemented on a custom event to allow [`EventListener`]s to convert [`PointerEvent`]s
+/// into the custom event type.
+pub trait EventFrom: Event {
```
colepoirier, line 38:
```suggestion
pub trait ForwardedEvent: Event {
```

crates/bevy_picking_core/src/output.rs
```diff
@@ -0,0 +1,628 @@
...
+impl<E: IsPointerEvent> EventListener<E> {
+    /// Create an [`EventListener`] that will run the supplied `on_event` function with access to
+    /// bevy [`Commands`].
+    pub fn new_run_commands(on_event: fn(&mut Commands, &mut EventData<E>)) -> Self {
+        Self { on_event }
+    }
+
+    /// Create an [`EventListener`] that will send an event of type `F` when the listener is
+    /// triggered, then continue to bubble the original event up this entity's hierarchy.
+    pub fn new_forward_event<F: EventFrom>() -> Self {
```
colepoirier, line 61:
```suggestion
    pub fn new_forward_event<F: ForwardedEvent>() -> Self {
```

crates/bevy_picking_core/src/output.rs
```diff
@@ -0,0 +1,628 @@
...
+            },
+        }
+    }
+
+    /// Create an [`EventListener`] that will send an event of type `F` when the listener is
+    /// triggered, then halt bubbling, preventing event listeners of the same type from triggering
+    /// on parents of this entity.
+    ///
+    /// Prefer using `new_forward_event` instead, unless you have a good reason to halt bubbling.
+    pub fn new_forward_event_and_halt<F: EventFrom>() -> Self {
```
colepoirier, line 78:
```suggestion
    pub fn new_forward_event_and_halt<F: ForwardedEvent>() -> Self {
```

crates/bevy_picking_core/src/output.rs
```diff
@@ -0,0 +1,628 @@
...
+}
+
+/// Extends the [`EntityCommands`] trait, allowing you to call these methods when spawning an
+/// entity.
+///
+/// # Usage
+///
+/// ```
+/// # struct MyForwardedEvent;
+/// # impl EventFrom for MyForwardedEvent {
```
colepoirier, line 99:
```suggestion
/// # impl ForwardedEvent for MyForwardedEvent {
```

crates/bevy_picking_core/src/output.rs
```diff
@@ -0,0 +1,628 @@
...
+/// # }
+/// # fn(mut commands: Commands){
+/// commands
+///     .spawn()
+///     .forward_events::<PointerClick, MyForwardedEvent>();
+/// # }
+/// ```
+pub trait EventListenerCommands {
+    /// Listens for events of type `E`. When found, an event of type `F` will be sent.
+    fn forward_events<E: IsPointerEvent, F: EventFrom>(&mut self) -> &mut Self;
```
colepoirier, line 112:
```suggestion
    fn forward_events<E: IsPointerEvent, F: ForwardedEvent>(&mut self) -> &mut Self;
```

crates/bevy_picking_core/src/output.rs
```diff
@@ -0,0 +1,628 @@
...
+    /// Listens for events of type `E`. When found, an event of type `F` will be sent. Finally,
+    /// bubbling will be halted. See [`event_bubbling`] for details on how bubbling works.
+    ///
+    /// Prefer using `forward_events` instead, unless you have a good reason to halt bubbling.
+    fn forward_events_and_halt<E: IsPointerEvent, F: EventFrom>(&mut self) -> &mut Self;
```
colepoirier, line 117:
```suggestion
    fn forward_events_and_halt<E: IsPointerEvent, F: ForwardedEvent>(&mut self) -> &mut Self;
```

crates/bevy_picking_core/src/output.rs
```diff
@@ -0,0 +1,628 @@
...
+}
+
+impl<'w, 's, 'a> EventListenerCommands for EntityCommands<'w, 's, 'a> {
+    fn forward_events<E: IsPointerEvent, F: EventFrom>(&mut self) -> &mut Self {
```
colepoirier, line 121:
```suggestion
    fn forward_events<E: IsPointerEvent, F: ForwardedEvent>(&mut self) -> &mut Self {
```

crates/bevy_picking_core/src/output.rs
```diff
@@ -0,0 +1,628 @@
...
+        self.commands().add(|world: &mut World| {
+            world.init_resource::<Events<F>>();
+        });
+        self.insert(EventListener::<E>::new_forward_event::<F>());
+        self
+    }
+    fn forward_events_and_halt<E: IsPointerEvent, F: EventFrom>(&mut self) -> &mut Self {
```
colepoirier, line 128:
```suggestion
    fn forward_events_and_halt<E: IsPointerEvent, F: ForwardedEvent>(&mut self) -> &mut Self {
```

crates/bevy_picking_selection/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+/// event when picked. This is useful for gizmos or other pickable UI entities.
+#[derive(Component, Debug, Copy, Clone)]
+pub struct NoDeselect;
+
+/// Unsurprising default multiselect inputs: both  control and shift keys.
+pub fn multiselect_events(
+    keyboard: Res<Input<KeyCode>>,
+    mut pointer_query: Query<&mut PointerMultiselect>,
+) {
+    let is_multiselect_pressed = keyboard.any_pressed([
```
colepoirier, line 76: How would a user provide custom multiselect_pressed KeyCodes?

colepoirier (changes requested):
Looks good to me! I only left a few type fixes and a couple clarifying questions. I'll probably have more questions when I'm trying to use it in my app, so I'll be sure to let you know when they crop up.

crates/bevy_picking_selection/src/lib.rs
aevyrie: The same way it's provided in this system, they would just write their own version. How do you suggest this is communicated?

colepoirier: Would the user have to write they own Selection plugin then? Would it make sense to have a way to specify the multislect keys when adding the plugin? Sorry if this doesn’t make sense, it’s really hot where I am so my brain is functioning poorly 😅

aevyrie: They can just reuse the selection plugin, but replace this system with their own. :)

You raise a good point though, it would be a good idea to extract this out as its own function so it's even easier to disable.

examples/bubbling.rs
aevyrie: I've added a bunch of comments, as well as a more advanced `generic_events` example. Let me know what you think.

aevyrie: Also, I'm not sure if the generic events are a feature looking for a problem, or if this is genuinely useful.

crates/bevy_picking_selection/src/lib.rs
colepoirier: Good point.

The ability to disable is good, though I’m thinking that making multiselect configurable as well as disable-able would be ideal. I’m not sure how this would be accomplished, perhaps @alice-i-cecile has some insight as this problem reminds my of her Leafwing Input Manager’s ‘action binding’ solution.

aevyrie: All inputs (even pointer clicks and movement) are exposed in this refactor, and the default bindings are easy to omit. This was done intentionally to support rebinding. This might make for a good example.

alice-i-cecile: Yeah; I'd be tempted to do a LWIM + bevy_mod_picking rebinding example. dev-dependencies are great for that.

colepoirier: > All inputs (even pointer clicks and movement) are exposed in this refactor, and the default bindings are easy to omit. This was done intentionally to support rebinding. This might make for a good example.

Awesome. Thanks for the clarification.

crates/bevy_picking_input/src/inputs.rs
```diff
@@ -0,0 +1,43 @@
...
+    keyboard: Res<Input<KeyCode>>,
+    touches: Res<Touches>,
+    mut cursor_query: Query<(&CursorId, &mut CursorMultiSelect, &mut CursorClick)>,
+) {
+    let is_multiselect_down = keyboard.any_pressed([
+        KeyCode::LControl,
+        KeyCode::RControl,
+        KeyCode::LShift,
+        KeyCode::RShift,
+    ]);
```
Weibye, line 16: This would likely benefit from having some layer of separation between the actions taken here (multiselect) and keyboard buttons. I'm sure you've considered that already and it would be a discussion about improving the bevy_input rather than doing anything here

crates/bevy_picking_input/src/mouse.rs
```diff
@@ -0,0 +1,46 @@
...
+            },
+        ));
+    }
+
+    for input in mouse_inputs.iter() {
+        let button = match input.button {
+            MouseButton::Left => PointerButton::Primary,
+            MouseButton::Right => PointerButton::Secondary,
+            MouseButton::Middle => PointerButton::Middle,
+            MouseButton::Other(_) => continue,
```
Weibye, line 34: This also feels like it should belong somewhere in bevy_input

crates/bevy_picking_input/src/touch.rs
```diff
@@ -0,0 +1,38 @@
...
+pub fn touch_pick_events(
+    touches: Res<Touches>,
+    mut pointer_moves: EventWriter<InputMove>,
+    mut pointer_clicks: EventWriter<InputPress>,
+) {
+    for touch in touches.iter() {
+        if touch.distance() != Vec2::ZERO {
+            let id = PointerId::Touch(touch.id());
+            let location = Location {
+                target: RenderTarget::Window(WindowId::primary()),
```
Weibye, line 18: Guessing this does not support touch input across multiple windows just yet?

crates/bevy_picking_selection/src/lib.rs
Weibye: Could keybinds be supplied in a plugin configuration resource somewhere? That you expose what _actions_ can be taken and supply a default keybind that the user can override if they want to bind a different key to an action?

examples/rapier.rs
```diff
@@ -49,10 +55,16 @@ fn setup(
...
+
+    // camera
     commands
-        .spawn_bundle(PerspectiveCameraBundle {
+        .spawn_bundle(Camera3dBundle {
             transform: Transform::from_xyz(-2.0, 2.5, 5.0).looking_at(Vec3::ZERO, Vec3::Y),
+            // projection: bevy::render::camera::Projection::Orthographic(OrthographicProjection {
+            //     scale: 0.01,
+            //     ..Default::default()
+            // }),
```
Weibye, line 66: Forgotten cleanup?

Weibye (commented):
This is a good piece of software, I appreciate the effort that went into this! I like the separation from the inputs and the backends which makes for a rather robust and modular kit. 

Random unsolicited thoughts and questions:

- Are there plans to upstream all or parts of this to Bevy?
- Would it be relatively easy to extended this in the future support VR raycast interactions?
- What about gamepad support?
- Most if not all of `crates/bevy_picking_core/src/pointer.rs` seems like it make sense as a part of bevy's input-system (given that were to be improved accordingly)

Weibye:
Overall, from a system architecture perspective and as a next step I would try to remove the need for hardcoded key-codes and provide some configurable way for the user to override keybindings that does not require having to implement the same systems again. But I suspect most of the heavy lifting should be done on the `bevy_input` side of things to enable use cases like this.

examples/bubbling.rs
```diff
@@ -0,0 +1,123 @@
...
+            material: materials.add(Color::WHITE.into()),
+            ..Default::default()
+        })
+        .insert_bundle(PickableBundle::default())
+        .insert(PickRaycastTarget::default())
+        // Because event forwarding can rely on event bubbling, events that target children of the
+        // parent cube will bubble up to this level and will fire off a `GreetMe` or `DeleteMe`
+        // event, depending on the event that bubbled up:
+        .forward_events::<PointerClick, DeleteMe>()
+        .forward_events::<PointerOver, GreetMe>()
```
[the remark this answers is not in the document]
aevyrie: What gets deleted is up to you. If you delete the listener, and the listener is at the top of the hierarchy, then yes. You could instead delete the target (which is what we do in this example), or you could put the listener somewhere else in the hierarchy.

crates/bevy_picking_core/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+            )
+            .add_system_set_to_stage(
+                CoreStage::First,
+                SystemSet::new()
+                    .after(PickStage::Focus)
+                    .label(PickStage::EventListeners)
+                    .with_system(event_bubbling::<output::Over>)
+                    .with_system(event_bubbling::<output::Out>)
+                    // PointerEnter does not allow bubbling
+                    // PointerLeave does not allow bubbling
```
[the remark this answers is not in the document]
aevyrie: I'm mirroring the behavior of the JS pointer event API, though these two events haven't been implemented yet. I'm unsure if the utility of the events is worth making them exceptions.

crates/bevy_picking_input/src/inputs.rs
aevyrie: All the systems that provide default inputs are just that - defaults. The API is designed so you can trigger any input with anything you want. I'll be making an example with leafwing input manager.

crates/bevy_picking_input/src/touch.rs
aevyrie: Ah good catch.

examples/rapier.rs
aevyrie: I left this here to make it easy to test ortho vs. perspective. Should at least leave a comment to that effect.

aevyrie:
> Overall, from a system architecture perspective and as a next step I would try to remove the need for hardcoded key-codes and provide some configurable way for the user to override keybindings that does not require having to implement the same systems again. But I suspect most of the heavy lifting should be done on the `bevy_input` side of things to enable use cases like this.

I thought this separation was clear, but it sounds like that is not the case. I put all default inputs into separate systems so they could be easily replaced. Perhaps it would help to consolidate all inputs into a single place in the root crate?

examples/rapier.rs
Weibye: Ah yeah, I see that now

aevyrie:
@Weibye thank you for your feedback!

>   Are there plans to upstream all or parts of this to Bevy?

Yes, that is the goal!

>   Would it be relatively easy to extended this in the future support VR raycast interactions?

Can you expand on this? There's no reason you can't create arbitrary pointers with arbitrary backends.

>   What about gamepad support?

Yes, that is the intent for all inputs being removable, and one of the motivators for adding custom cursors not tied to any hardware, you can definitely control a pointer with a gamepad. I'll probably make a rebinding example that shows this.

>   Most if not all of crates/bevy_picking_core/src/pointer.rs seems like it make sense as a part of bevy's input-system (given that were to be improved accordingly)

Yup, that's what I'm hoping to do here. I need to make a picking backend for bevy_ui, before it's even possible though.

crates/bevy_picking_selection/src/lib.rs
aevyrie: > Could keybinds be supplied in a plugin configuration resource somewhere? 
I'm hesitant to do this because I don't want to make any assumptions about how inputs work. E.g. I don't want to constrain an input to only work with the keyboard or mouse. That's why all input events are just bevy events or resources, to allow users to update these inputs with whatever they want (gamepads, joysticks, the current temperature in Belize, a web request, etc. :) )

Weibye:
> I thought this separation was clear, but it sounds like that is not the case. I put all default inputs into separate systems so they could be easily replaced. Perhaps it would help to consolidate all inputs into a single place in the root crate?

It could just be that my gut reaction is that "unless it is _the_ system responsible for receiving input into the game / app, there should not be direct mentions of actual key codes", instead there should be something defining an _action_ of what can be triggered then it is up to the input manager to bind buttons to actions. 

My background here is Unity and for quite some time it was idiomatic to reference buttons directly in gameplay code. While _easy_ it caused maintenance nightmare. Then a while a ago came the new input manager provided a separation between physical input and in-game actions which was _heavenly_ and made everything much easier to maintain. (The actual system is poorly made and cumbersome to work with, but the concept is great)

Also, make sure the feedback makes sense to you before you act on it, I'm just a person on the internet 😃

Weibye:
> > Would it be relatively easy to extended this in the future support VR raycast interactions?
> 
> Can you expand on this? There's no reason you can't create arbitrary pointers with arbitrary backends.

I think this is the answer then :) A VR input ray, (to interact with UI panels from a distance, like a laser-pen) would simply be another pointer input

bonsairobo (commented):
I'm taking a break. I've gotten through everything except for:

- `bevy_picking_input`
- `bevy_picking_selection`
- examples

bonsairobo (commented):
Got through input and selection. Will run the examples next time.

backends/bevy_picking_shader/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+/// Commonly used imports for the [`bevy_picking_raycast`] crate.
+pub mod prelude {
+    // pub use crate::;
+}
+
+/// Adds support for shader picking to `bevy_mod_picking`.
+pub struct ShaderPlugin;
+impl Plugin for ShaderPlugin {
+    fn build(&self, _app: &mut App) {
+        unimplemented!();
```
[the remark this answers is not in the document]
aevyrie: I've done some high level investigation, and I'm not sure how to best implement a shader backend yet. Ideally I could just mark entities with a component and render them to an offscreen buffer, but it doesn't seem like I have the flexibility I would need to modify the render graph to make this work. Rob Swain suggested I make my own version of `StandardMaterial`, but I worry this won't be flexible enough.

My intent was to include this feature, but I'll probably end up deferring it to a subsequent PR.

crates/bevy_picking_core/src/output.rs
```diff
@@ -0,0 +1,646 @@
...
+}
+impl Bubble {
+    /// Stop this event from bubbling to the next parent.
+    pub fn burst(&mut self) {
+        *self = Bubble::Burst;
+    }
+}
+
+/// This trait restricts the types of events that can be used as [`PointerEvent`]s.
+pub trait IsPointerEvent: Send + Sync + 'static + Display + Clone {
```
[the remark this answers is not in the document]
aevyrie: IIRC, I needed this for making it compatible with events, will need to double check.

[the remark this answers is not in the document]
bonsairobo: Oh I just meant the placement of the lifetime between traits, as opposed to putting `'static` at the beginning or end.

crates/bevy_picking_input/src/inputs.rs
```diff
@@ -0,0 +1,43 @@
...
+                if pressed && !click.as_ref().is_clicked {
+                    click.is_clicked = true;
+                } else if !pressed && click.as_ref().is_clicked {
+                    click.is_clicked = false;
+                }
+            }
+            _ => (),
+        }
+    }
+}
```
bonsairobo, line 43: I just realized this `inputs.rs` file is not being used anywhere (no `inputs` module is declared). So I don't think it compiles as-is.

examples/bubbling.rs
```diff
@@ -0,0 +1,127 @@
...
+struct GreetMe(Entity);
+impl ForwardedEvent for GreetMe {
+    fn from_data<E: IsPointerEvent>(event_data: &PointerEventData<E>) -> GreetMe {
+        GreetMe(event_data.target())
+    }
+}
+impl GreetMe {
+    fn handle_events(mut greet: EventReader<GreetMe>) {
+        for event in greet.iter() {
+            info!("Hello {:?}!", event.0);
```
bonsairobo, line 75: Might be nice to print both the target and listener entities so it's more obvious where the events are coming from.

bonsairobo:
I tested out all of the examples on this branch. They behaved as expected. :tada:

crates/bevy_picking_core/src/output.rs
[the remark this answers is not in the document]
aevyrie: Doh, thanks for clarifying. :)

crates/bevy_picking_core/src/focus.rs
```diff
@@ -0,0 +1,118 @@
...
+
+/// Maps [`RenderLayers`] to the map of entities within that pick layer, sorted by depth.
+type LayerMap = BTreeMap<Layer, DepthMap>;
+
+/// Maps Pointers to a [`LayerMap`]. Note this is much more complex than the [`HoverMap`] because
+/// this data structure is used to sort entities by layer then depth for every pointer.
+type OverMap = HashMap<PointerId, LayerMap>;
+
+/// Maps pointers to the entities they are hovering over. "Hovering" refers to the actual hover
+/// interaction state, in this case, not just whether the pointer happens to be over the entity.
```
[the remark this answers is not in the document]
aevyrie: "Over" = the pointer passes a hit test, and is within the bounds of the object
"Hover" = the pointer is over the object AND the pointer has done a depth test to see if this object should be in its hover state. The depth test looks at the topmost object under the cursor, and checks iteratively to see if the pointer should "pass through" to the next entity, stopping when it reaches the first object it can interact with.

Ex: UI drawn over a viewport. The pointer is "over" a UI button and a mesh in the scene. However, only the UI button will be in the "hover" state.

aevyrie:
> I tested out all of the examples on this branch. They behaved as expected. tada

@bonsairobo any thoughts on what would make for some good examples? I'd like to add examples that demonstrate how you can make use of the high level pointer events to do complex things easily. Maybe a drag and drop example?

crates/bevy_picking_core/src/focus.rs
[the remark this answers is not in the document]
bonsairobo: Yea that makes complete sense. I realized this after reading the systems that build the `HoverMap`. It might not be necessary to add this much detail to _this_ doc string, but maybe just a short explanation of what "hover interaction" actually means would be nice.

bonsairobo:
> any thoughts on what would make for some good examples? I'd like to add examples that demonstrate how you can make use of the high level pointer events and listeners to do complex things easily. Maybe a drag and drop example?

@aevyrie Definitely drag and drop. I noticed that was missing.

Let me tell you about one time I built a UI with drag-and-drop (using Amethyst) :). I basically had a little table of square cells where each row held the timeline of actions for a character. Then you could create a plan by placing token entities (each representing a type of action) into the cells.

This was interesting as a UI because it involved drag/hover/drop events, and it also had to enforce limits on what could be placed in a cell, which required noticing a problem and then animating the token entity back into a valid spot. You could also notice when a token is being dragged over a cell, and change the highlight of the cell to indicate valid/invalid placement. I think this example would require use of the high-level pointer events, event forwarding, and listeners.

bonsairobo:
I guess I basically just described the feature set of pretty much any card game.

crates/bevy_picking_core/src/output.rs
```diff
@@ -593,25 +618,25 @@ pub fn send_drag_over_events(
 ) {
     // Fire PointerDragEnter events.
     for over_event in pointer_over.iter() {
-        if let Some(Some(dragged)) = drag_map.get(&over_event.id()) {
-            if &over_event.target() != dragged {
+        if let Some(&Some(dragged)) = drag_map.get(&over_event.id()).to_owned() {
```
bonsairobo, line 621: Why `to_owned`?

backends/bevy_picking_ui/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+            CoreStage::First,
+            SystemSet::new()
+                .label(PickStage::Backend)
+                .with_system(ui_picking),
+        );
+    }
+}
+
+/// Computes the UI node entities under each pointer
+pub fn ui_picking(
```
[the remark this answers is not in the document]
aevyrie: Yes, this is useful for system ordering, so users can refer to it as `.after(ui_Picking)`. I also don't see the harm in at least making it callable if someone wants to reorder systems manually.

crates/bevy_picking_core/src/focus.rs
```diff
@@ -0,0 +1,118 @@
...
+        };
+        for over in event.over_list.iter() {
+            let layer: Layer = render_layers
+                .get(over.entity)
+                .ok()
+                .map(|layer| layer.iter().max())
+                .flatten()
+                .unwrap_or(0);
+
+            layer_map.entry(layer).or_insert_with(BTreeMap::new);
```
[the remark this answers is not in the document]
aevyrie: I don't follow, it's `get_mut` because I need a mutable reference to the depth map inside the layer map.

crates/bevy_picking_core/src/focus.rs
```diff
@@ -0,0 +1,118 @@
...
+    build_pointer_map(render_layers, &mut under_pointer, &mut pointer_over_map);
+    build_hover_map(&pointers, focus, pointer_over_map, &mut hover_map);
+}
+
+/// Clear non-empty local maps, reusing allocated memory.
+fn reset_local_maps(hover_map: &mut ResMut<HoverMap>, pointer_over_map: &mut Local<OverMap>) {
+    for entity_set in hover_map.values_mut() {
+        if !entity_set.is_empty() {
+            entity_set.clear()
+        }
```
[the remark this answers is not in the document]
aevyrie: I was thinking the cost of clearing unnecessarily would outweigh the branching. I guess only benchmarking would tell.

crates/bevy_picking_core/src/focus.rs
```diff
@@ -0,0 +1,118 @@
...
+    render::view::{Layer, RenderLayers},
+    ui::FocusPolicy,
+    utils::{FloatOrd, HashMap, HashSet},
+};
+
+/// A map of entities sorted by depth.
+type DepthMap = BTreeMap<FloatOrd, Entity>;
+
+/// Maps [`RenderLayers`] to the map of entities within that pick layer, sorted by depth.
+type LayerMap = BTreeMap<Layer, DepthMap>;
```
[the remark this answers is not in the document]
aevyrie: Hmm, I think it's useful to have sets of sorted entities per layer. We might want to add some per-layer functionality, and this structure makes that easier. It's also easier to reason about, I think, which was one of the goals of this section.

crates/bevy_picking_core/src/focus.rs
aevyrie: I'm not quite sure what to add here, the function is only three lines long, and each of those function document what they do. I'd like to avoid duplicating information. I added another line of context to the comment.

crates/bevy_picking_core/src/output.rs
```diff
@@ -0,0 +1,646 @@
...
+        while let Ok((event_listener, parent)) = listeners.get(listener) {
+            if let Some(event_listener) = event_listener {
+                let event_data = PointerEventData {
+                    id: event.id,
+                    listener,
+                    target: event.target,
+                    event: event.event.clone(),
+                };
+                let mut bubble = Bubble::default();
+                let callback = event_listener.callback;
```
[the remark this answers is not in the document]
aevyrie: Not strictly, no, however I find the function pointer calling syntax quite weird:

```rs
                (event_listener.callback)(&mut commands, &event_data, &mut bubble);
```
So this is purely for readability that gets optimized away.

crates/bevy_picking_core/src/output.rs
```diff
@@ -0,0 +1,646 @@
...
+    event: &PointerEvent<E>,
+    new_interaction: Interaction,
+    pointer_interactions: &mut Query<(&PointerId, &mut PointerInteraction)>,
+    entity_interactions: &mut Query<&mut Interaction>,
+) {
+    if let Some(mut interaction_map) = pointer_interactions
+        .iter_mut()
+        .find_map(|(id, interaction)| (*id == event.id).then_some(interaction))
+    {
+        interaction_map.insert(event.target, new_interaction);
```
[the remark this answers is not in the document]
aevyrie: No, and I don't believe anything needs to. It's a lazily-initialized list of interactionss, and is also used to check how the state has changed. I could maybe remove items if they haven't changed for a few frames?

crates/bevy_picking_core/src/output.rs
```diff
@@ -572,22 +580,30 @@ pub fn send_click_and_drag_events(
             }
         }
     }
-
+j
```
MitchellHansen, line 583: i think you dropped a j here

crates/bevy_picking_core/src/focus.rs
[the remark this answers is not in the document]
bonsairobo: There is almost definitely branching inside of `clear` to handle this anyway.

crates/bevy_picking_core/src/focus.rs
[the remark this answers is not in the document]
bonsairobo: I'm saying you could do:

```rust
let depth_map = layer_map.entry(layer).or_insert_with(BTreeMap::new);
```

crates/bevy_picking_core/src/output.rs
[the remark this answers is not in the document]
bonsairobo: It's basically a memory leak for long-running processes.

crates/bevy_picking_core/src/focus.rs
tbarusseau: Wow same! Currently using this for a level editor coupled with `bevy_egui` and... Yup I could use this 😄

PR #146, opened 2022-07-26
Make highlights optional

mxgrey:
Thanks for making this picking library; it's super easy to use and valuable!

I found that in my use case the hover highlight was very distracting. Colors would flash wildly just from moving the cursor across the window.

This PR reworks highlighting behavior so that all the global highlights are optional. When a global highlight is `None` and the entity highlight component is also `None`, then there will simply be no highlight applied (..except in the case of Hover+Selected explained below..).

This also adds a `hover_selected` field which will be used when an object is both selected and being hovered over at the same time. If `hover_selected` is `None` for both the entity highlighting and the global highlighting, then the coloring falls back to hover and selected behavior with this prioritization:
```
entity hover_selected > global hover_selected > entity hovered > entity selected > global hovered > global selected > no highlight
```

I based this PR off of the `v0.7.0` tag because `main` wasn't able to compile for me and I couldn't be bothered to figure out why :grimacing: ...But if you'd like me to revert the `Cargo.toml` and/or the `README.md` just let me know.

Unfortunately this is a bit of an API break so I can understand if that may be controversial. Luckily all of the examples continue to work as-is. The change will only impact users who are customizing their default coloring, and they only need to deal with the switch from `T` to `Option<T>` which shouldn't be a very big deal.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bevy_mod_picking"
-version = "0.6.0"
+version = "0.6.1"
 authors = ["Aevyrie <aevyrie@gmail.com>"]
 edition = "2021"
 license = "MIT"
@@ -11,13 +11,11 @@ categories = ["game-engines", "rendering"]
 resolver = "2"
 
 [dependencies]
-bevy = { git = "https://github.com/bevyengine/bevy", branch = "main", default-features = false, features = [
-    "render",
-] }
-bevy_mod_raycast = { git = "https://github.com/aevyrie/bevy_mod_raycast", branch = "main" }
+bevy = { version = "0.7", default-features = false, features = ["render"] }
+bevy_mod_raycast = "0.4.0"
 
 [dev-dependencies]
-bevy = { git = "https://github.com/bevyengine/bevy", branch = "main", default-features = false, features = [
+bevy = { version = "0.7", default-features = false, features = [
     "bevy_winit",
     "x11",
 ] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ resolver = "2"
 
 [dependencies]
 bevy = { version = "0.7", default-features = false, features = ["render"] }
-bevy_mod_raycast = "0.4.0"
+bevy_mod_raycast = "0.5.0"
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


aevyrie:
Hey @mxgrey, thanks for the contribution. Does disabling the highlighting plugin not achieve what you are looking for?

Additionally, you will probably want to base your changes on #137, this substantially changes, well, everything.

mxgrey:
Oh, I didn't notice #137 before opening the PR! I'll take a look at that and see if there's a way to achieve what I'm hoping for in there.

> Does disabling the highlighting plugin not achieve what you are looking for?

I'd like to keep the highlighting plugin running because I want to keep using the selection highlighting feature (i.e. users can select objects by clicking on them). I just don't want colors to change due to hovering. In fact the next thing I'm planning is to piggyback on your plugin's `PickingCamera` component to add a little marker at the points that are being hovered over as a less aggressive alternative to changing the color of the whole mesh. Unless I'm misunderstanding your suggestion, I think I need to keep the plugin enabled for all of that to work.

mxgrey:
It looks like it would be pretty easy to migrate my changes onto your big refactor by changing [these fields](https://github.com/aevyrie/bevy_mod_picking/pull/137/files#diff-13a546f2a34c4d3f2eedd72db35b66e0cd40069bf2672a013c56e301fa3ef95eR103-R112) to optional and then tweaking the logic in [`update_highlight_asset`](https://github.com/aevyrie/bevy_mod_picking/pull/137/files#diff-13a546f2a34c4d3f2eedd72db35b66e0cd40069bf2672a013c56e301fa3ef95eR179).

Does the behavior that I'm proposing sound reasonable enough to include it in your new refactor? If so I'm happy to open a PR targeting that.

aevyrie:
I *think* so. I'd have to consider the consequences of this. I do want to support this use case!

mxgrey:
Maybe I can add an example that makes use of hover being disabled. At the very least you may want to try out the existing example and see how you like the `hover_selected` behavior. I think it makes the user experience a lot smoother when the hover highlight isn't taking precedence over the selected highlight.

aevyrie:
Closing this PR. Willing to accommodate this use case, however this PR is very out of date w.r.t. the beta branch.

Issue #138, opened 2022-06-06
Relicense bevy_mod_picking under dual MIT/Apache-2.0


aevyrie:
**To agree to this relicense, please read the details in this issue, then leave a comment with the following message:**

`I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to choose either at their option.`

If you disagree, please respond with your reasoning (just don't expect us to change course at this point). Anyone who doesn't agree to the relicense will have any contributions that qualify as "copyrightable" removed or re-implemented.

Why: https://github.com/bevyengine/bevy/issues/2373

# What will this look like?

After getting explicit approval from each contributor of copyrightable work
(as not all contributions qualify for copyright, due to not being a "creative
work", e.g. a typo fix), we will add the following file to our README:

### License

Licensed under either of

 * Apache License, Version 2.0, ([LICENSE-APACHE](LICENSE-APACHE) or http://www.apache.org/licenses/LICENSE-2.0)
 * MIT license ([LICENSE-MIT](LICENSE-MIT) or http://opensource.org/licenses/MIT)

at your option.

### Contribution

Unless you explicitly state otherwise, any contribution intentionally submitted
for inclusion in the work by you, as defined in the Apache-2.0 license, shall be dual licensed as above, without any
additional terms or conditions.

We will add LICENSE-{MIT,APACHE} files containing the text of each license. We will also update the license metadata in our Cargo.toml to:

license = "MIT OR Apache-2.0"

Ratysz:
I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to choose either at their option.

obsgolem:
I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to choose either at their option.

gavlig:
I license past and future contributions to bevy_mod_picking under the dual MIT/Apache-2.0 license, allowing licensees to choose either at their option.

(I specified the project name in text, hope that's alright)

YoshieraHuang:
I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to choose either at their option.

mcobzarenco:
I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to choose either at their option.

thlorenz:
I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to choose either at their option.

BrunoWallner:
[...]

PR #168, opened 2022-11-07
Update to Bevy 0.9

olegomon:
@aevyrie I prepared a small PR which updates some of the deprecated Bevy APIs. Let me know if that's ok, any feedback appreciated.

- use spawn() instead of spawn_bundle()
- use insert() instead of insert_bundle()
- use new DefaultPlugins API

I split the `HighlightablePickingPlugins` from being a part of `DefaultPickingPlugins` I was not sure what would be a better alternative to merging two plugin groups in the PluginGroup::build trait implementation. Is that ok or do you have a suggestion on what to do there?

### DIFF
--- a/examples/deselection.rs
+++ b/examples/deselection.rs
@@ -1,16 +1,19 @@
 use bevy::{prelude::*, window::PresentMode};
-use bevy_mod_picking::{DefaultPickingPlugins, NoDeselect, PickableBundle, PickingCameraBundle};
+use bevy_mod_picking::{HighlightablePickingPlugins, DefaultPickingPlugins, NoDeselect, PickableBundle, PickingCameraBundle};
 
 /// This example is identical to the 3d_scene example, except a cube has been added, that when
 /// clicked on, won't deselect everything else you have selected.
 fn main() {
     App::new()
-        .insert_resource(WindowDescriptor {
-            present_mode: PresentMode::AutoNoVsync, // Reduce input latency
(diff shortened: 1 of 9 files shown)


examples/deselection.rs
```diff
@@ -23,34 +26,34 @@ fn setup(
...
     // plane
     commands
-        .spawn_bundle(PbrBundle {
+        .spawn(PbrBundle {
             mesh: meshes.add(Mesh::from(shape::Plane { size: 5.0 })),
             material: materials.add(Color::rgb(0.3, 0.5, 0.3).into()),
             ..Default::default()
         })
-        .insert_bundle(PickableBundle::default());
+        .insert(PickableBundle::default());
```
lepsalex, line 34: consolidating as many operations as possible to improve ergonomics and cut down on archetype moves is also recommended now, ex:

```rust
commands.spawn((
        PbrBundle {
            mesh: meshes.add(Mesh::from(shape::Plane { size: 5.0 })),
            material: materials.add(Color::rgb(0.3, 0.5, 0.3).into()),
            ..Default::default()
        },
        PickableBundle::default(),
    ));
```

aevyrie:
Thank you!

aevyrie:
@olegomon could you please read this https://github.com/aevyrie/bevy_mod_picking/issues/138 and comment on the thread so we can relicense as MIT + Apache?

PR #175, opened 2023-01-08
Use `insert` instead of deprecated `insert_bundle`

enbugger:
As per Bevy 0.9
```
    #[deprecated(
        since = "0.9.0",
        note = "Use `insert` instead, which now accepts bundles, components, and tuples of bundles and components."
    )]
    pub fn insert_bundle(&mut self, bundle: impl Bundle) -> &mut Self {
        self.insert(bundle)
    }
```

Luckily this affects only documentation


aevyrie (approved):
Thanks!

PR #177, opened 2023-01-10
Allow a camera with a RenderTarget::Image to set the mouse position.

mattatz:
Thanks for developing the bevy_mod_picking library!

I am developing a bevy application using post effect in the wasm environment, and I have found a problem where I cannot get the mouse position, which is the starting position of the ray, when the camera that is the picking source has RenderTarget::Image.

For now, this pull request simply allows the mouse position to be set in `get_inputs`, but should this modification only take effect in wasm builds?

I would be happy to have it merged if this is not a problem.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bevy_mod_picking"
-version = "0.9.0"
+version = "0.10.0"
 authors = ["Aevyrie <aevyrie@gmail.com>"]
 edition = "2021"
 license = "MIT"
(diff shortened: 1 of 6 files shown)


aevyrie:
Thanks for the contribution, I need to think about this some more. FYI, most development efforts are probably best targeted toward the `beta` branch.

mattatz:
@aevyrie 
Hi aevyrie!
I have been able to confirm that there are significant specification changes in the `beta` branch.
I would like to create a new branch from the `beta` branch, so I will close this PR for the time being.

PR #180, opened 2023-01-27
Plugin Rewrite Beta

aevyrie:


bonsairobo:
If it's not too much work, I'd like to get better support for using the depth and surface normal of a ray-entity intersection. Right now the depth is computed by a picking backend in order to choose the entity closest to the camera. But this depth is not easily accessible by picking event handlers. It would be nice to somehow reuse the backend calculations by storing the intersection metadata (depth + surface normal) either in a component or the event itself.

bonsairobo:
Or maybe it would be more idiomatic to replace `HashMap<PointerId, T>` with `T` components of the pointer entities?

src/debug.rs
```diff
@@ -0,0 +1,274 @@
...
+        Entity,
+        &pointer::PointerId,
+        &pointer::PointerLocation,
+        &pointer::PointerPress,
+        &events::PointerInteraction,
+    )>,
+    #[cfg(feature = "selection")] selection: Query<Option<&selection::PointerMultiselect>>,
+    primary_window: Query<&Window, With<PrimaryWindow>>,
+) {
+    let font = asset_server.load("fonts/FiraMono-Medium.ttf");
```
Vrixyz, line 217: This could be configurable by user ? (I'm fine with leaving it as default)

If every plugin hardcode their debug font it could lead to a lot of arbitrary debug fonts just sitting there for debug compatibility, not very practical.

assets/images/bavy.png
Vrixyz, line 1: I like the meme-ing :)

I was a bit confused to not see this one though:
bavy

I like yours! but should it be named something else ? ("bovy" ?) ; or could be a running joke where all "bavy" icons are slightly off.

src/debug.rs
aevyrie: I have no idea if there is a good solution to this, I've never seen one anyway. Suggestions welcome.

assets/images/bavy.png
aevyrie: I don't own the rights to either, so I just made my own. :)

src/debug.rs
aevyrie: Managed to get this working, though it's a bit cursed. https://discord.com/channels/691052431525675048/691052431974465548/1096216676460146758

backends/bevy_picking_sprite/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+        let over_list = sorted_sprites
+            .iter()
+            .copied()
+            .filter_map(
+                |(entity, sprite, image, global_transform, visibility, focus)| {
+                    if blocked || !visibility.is_visible() {
+                        return None;
+                    }
+
+                    blocked = focus != Some(&FocusPolicy::Pass);
```
Vrixyz, line 79: I think this should be checked only if `contains_cursor` is true.

backends/bevy_picking_sprite/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+
+                    let extents = sprite
+                        .custom_size
+                        .or_else(|| images.get(image).map(|f| f.size()))
+                        .map(|size| size / 2.0)?;
+
+                    let anchor_offset = sprite.anchor.as_vec() * extents;
+
+                    let target_size = if let Some(t) =
+                        location.target.get_render_target_info(&windows, &images)
```
Vrixyz, line 92: I think this is a "complicated" research of target size, half target size depending on where the Y is on the screen, not taking into account camera position... etc etc.

There is a bevy builtin way to get world position from a camera:
```rs
let Some(cursor_position) = camera
            .viewport_to_world(camera_transform, cursor_position)
            .and_then(|ray| Some(ray.get_point(0f32))) else {
                continue;
            };
```

Check it in "real code" there: https://github.com/Vrixyz/bevy_jam_3/blob/a58ce17b9f03b1ab5e36ac81acfa070576cc1d1f/src/picking_aabb.rs#L72 ; adapted from your sprite backend, hope you don't mind :)

aevyrie: I don't think this is applicable for the sprite case. There is no way to get the sprite bounds in world coordinates that I am aware of, so I have to compute it in screen space.

Vrixyz: Oh the bounds of a sprite is expressed in screen size ? Alright, I'm just worried this solution won't support camera position, rotation and scale/zoom ; anyway it's more an implementation thing and could be adressed easily after this big PR if need be 👍

backends/bevy_picking_sprite/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+            target / 2.0
+        } else {
+            continue;
+        };
+        let cursor_position = location.position;
+        let cursor_pos_centered = cursor_position - target_half_extents;
+
+        let picks: Vec<(Entity, PickData)> = sorted_sprites
+            .iter()
+            .copied()
```
Vrixyz, line 75: I'm not sure this copy is still needed ? As the filter_map will create new instances not referring to original data

backends/bevy_picking_sprite/src/lib.rs
aevyrie: > I'm just worried this solution won't support camera position, rotation and scale/zoom.

Ah! I misunderstood. I'm not very familiar with 2D, but it does seem like there should be a better way to convert a screen space coordinate to world space for 2D.

backends/bevy_picking_sprite/src/lib.rs
aevyrie: `iter()` yields references. The `copied` is needed to return owned data. The alternative is to dereference in the code that follows, but it will be copying regardless, which adds noise for no benefit. The `copied` makes the `Entity` and `Option<&FocusPolicy>` owned instead of references.

backends/bevy_picking_sprite/src/lib.rs
aevyrie: This should now be fixed in 53e4dfe2696822742f187e12165ea06eaad690e6, thanks for raising this issue!

PR #181, opened 2023-01-30
Use released version of bevy_rapier3d

nicopap:
Now that there is a published version of rapier compatible with bevy 0.9,
there is no reason to depend on ira's fork.


aevyrie:
Git refuses to merge these changes for some reason. I made a clean PR here, and gave you co-author credit: #186

PR #185, opened 2023-01-31
Debug background for debug info

Aceeri:
Hard to read on bright backgrounds atm, so switching to use the debug_text egui function that gives a semi-transparent black background to it


aevyrie:
Looks great. +1

PR #189, opened 2023-02-16
Forward PointerButton in events

Aceeri:
lets us filter events that we don't care about, e.g. I only care about the primary click in bevy_editor_pls for picking

### DIFF
--- a/crates/bevy_picking_core/src/output.rs
+++ b/crates/bevy_picking_core/src/output.rs
@@ -599,7 +599,7 @@ pub fn send_click_and_drag_events(
 ) {
     // Only triggers when over an entity
     for move_event in pointer_move.iter() {
-        for button in PointerButton::all_buttons() {
+        for button in PointerButton::iter() {
             let moving_pointer_is_down = matches!(
                 down_map.get(&(move_event.pointer_id(), button)),
                 Some(Some(_))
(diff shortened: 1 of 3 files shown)


crates/bevy_picking_core/src/output.rs
```diff
@@ -335,20 +335,29 @@ pub struct Out;
 pub type PointerDown = PointerEvent<Down>;
 #[derive(Copy, Clone, Eq, PartialEq, Debug, Reflect)]
 /// The inner [`PointerEvent`] type for [`PointerDown`].
-pub struct Down;
+pub struct Down {
+    /// Pointer button pressed to trigger this event.
+    pub button: PointerButton,
+}
 
 /// Fires when a the pointer primary button is released over the `target` entity.
```
aevyrie, line 343: This comment is incorrect - it fires when *any* pointer button is pressed, as you correctly pointed out. :)

aevyrie: Could you also correct the commend on pointerdown please?

aevyrie (changes requested):
Looks great. Could you also please add this on the `Drag` and `Drop` types to be consistent?

aevyrie (approved):

PR #191, opened 2023-03-08
Migrate to bevy 0.10

Fincap:
Hi all,

I've tried my hand at migrating this plugin to be compatible with the new Bevy 0.10 version. This is forked from the main branch and not the beta branch, with the intent of getting a working version of this plugin released ASAP.

This is currently in draft phase, as there are a couple outstanding issues. While the changes committed will compile fine, functionality when executing the examples is broken. I am creating this pull request now to seek assistance with fixing this issue before finalising.

I will leave annotations describing the issues requiring resolution. Any and all responses are appreciated :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ categories = ["game-engines", "rendering"]
 resolver = "2"
 
 [dependencies]
-bevy = { version = "0.10.0", default-features = false, features = ["bevy_render"] }
+bevy = { version = "0.10.0", default-features = false, features = ["bevy_render", "bevy_ui"] }
 
 bevy_mod_raycast = {git = "https://github.com/soerenmeier/bevy_mod_raycast", branch = "bevy-0.10"}
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,12 +11,12 @@ categories = ["game-engines", "rendering"]
 resolver = "2"
 
 [dependencies]
-bevy = { version = "0.9", default-features = false, features = ["render"] }
+bevy = { version = "0.10.0", default-features = false, features = ["bevy_render"] }
 
-bevy_mod_raycast = "0.7"
+bevy_mod_raycast = {git = "https://github.com/soerenmeier/bevy_mod_raycast", branch = "bevy-0.10"}
 
 [dev-dependencies]
-bevy = { version = "0.9", default-features = false, features = [
+bevy = { version = "0.10.0", default-features = false, features = [
     "bevy_pbr",
     "bevy_winit",
     "bevy_ui",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,9 +11,12 @@ categories = ["game-engines", "rendering"]
 resolver = "2"
 
 [dependencies]
-bevy = { version = "0.10.0", default-features = false, features = ["bevy_render", "bevy_ui"] }
+bevy = { version = "0.10.0", default-features = false, features = [
+    "bevy_render",
+    "bevy_ui",
+] }
 
-bevy_mod_raycast = {git = "https://github.com/soerenmeier/bevy_mod_raycast", branch = "bevy-0.10"}
+bevy_mod_raycast = "0.8"
 
 [dev-dependencies]
 bevy = { version = "0.10.0", default-features = false, features = [
--- a/src/mouse.rs
+++ b/src/mouse.rs
@@ -2,6 +2,7 @@ use crate::{PickingCamera, UpdatePicks};
 use bevy::{
     prelude::*,
     render::camera::{Camera, RenderTarget},
+    window::WindowRef,
 };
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -11,12 +11,12 @@ categories = ["game-engines", "rendering"]
 resolver = "2"
 
 [dependencies]
-bevy = { version = "0.9", default-features = false, features = ["render"] }
+bevy = { version = "0.10.0", default-features = false, features = ["bevy_render"] }
 
-bevy_mod_raycast = "0.7"
+bevy_mod_raycast = {git = "https://github.com/soerenmeier/bevy_mod_raycast", branch = "bevy-0.10"}
```
Fincap, line 16: Temporarily pointing `bevy_mod_raycast` dependency to a 0.10-compatible fork. There is an active pull request to merge this branch into the main repo, so this will need to be updated once that happens.

src/mouse.rs
```diff
@@ -57,8 +58,12 @@ fn get_inputs<'a>(
     let cursor_latest = match cursor.iter().last() {
         Some(cursor_moved) => {
             if let RenderTarget::Window(window) = camera.target {
-                if cursor_moved.id == window {
-                    Some(cursor_moved.position)
+                if let WindowRef::Entity(camera_entity) = window {
+                    if cursor_moved.window == camera_entity {
+                        Some(cursor_moved.position)
```
Fincap, line 63: This is the main section causing issues due to the changes bevy 0.10 made to how windows are handled. The `cursor_moved.window` handle points to the window entity, but extracting the `WindowRef` from the `camera_target` returns `WindowRef::Primary` instead of `WindowRef::Entity`.

Fincap:
Additionally, given the issues get resolved, should a commit be made to bump the version of the plugin to something like 0.11.1 as a "bridge" version that is compatible with the latest bevy version and is available to be used while the beta refactor is still undergoing completion?

src/mouse.rs
Fincap: I've tried removing the nested checks entirely, and just returning `Some(cursor_moved.position`. This actually worked, but the rays seemed to be casting _towards_ the camera from the world, and not from the camera like it's supposed to.

soerenmeier: You could query for the `PrimaryWindow` and then if `WindowRef::Primary` use that entity.
The problem about the rays being reversed, is caused by the default implementation of FocusPolicy which got changed in https://github.com/bevyengine/bevy/pull/7161.

Both are fixes in this commit: https://github.com/aevyrie/bevy_mod_picking/commit/aaf439f5ef645fae36e98969b898e52aa4b44812

Fincap:
Thanks for your fix @soerenmeier :) Once https://github.com/aevyrie/bevy_mod_raycast/pull/72 is approved, I will update the cargo file and take this PR out of draft mode.

Fincap:
I realised that removing many of the `PickingSystem` variants, while not impacting the library or examples, may inadvertently break module user's reliance on the labels of particular systems.

A result of this is that each plugin's systems are now individually labelled in `PickingSystem`, however three additional sets have been created (`PickingSet`, `InteractableSet`, and `CustomHighlightSet`) which represent the three "groups" of systems. They are used in the library to order each system and apply run conditions. Alternatively, these three variants can be removed and a `distributive_run_if` can be applied to the systems tuple, but that means that each system in the set is individually queried for its run condition each loop rather than one for the system set as a whole.

Cargo.toml
```diff
@@ -11,12 +11,12 @@ categories = ["game-engines", "rendering"]
 resolver = "2"
 
 [dependencies]
-bevy = { version = "0.9", default-features = false, features = ["render"] }
+bevy = { version = "0.10.0", default-features = false, features = ["bevy_render"] }
```
Rubixdarcy, line 14: @Fincap Does this also need feature `bevy_ui`? When I run cargo build I get errors because it depends on some UI features existing in the prelude.

Fincap: Good catch, thank you. I had only been testing with `cargo run`, but you're right `cargo build` fails unless `bevy_ui` is included, which is a bit surprising to me. I'm not overly familiar with cargo so I'm not certain why run works where build fails, I would be interested if anyone could provide insight on that.

The issue has been fixed.

Fincap:
> Any news about the upgrade?

I am still waiting for [bevy_mod_raycast](https://github.com/aevyrie/bevy_mod_raycast) to be updated to 0.10 before moving this PR out of draft. 

In the meantime, if you'd like to use this branch in your project you can update your `Cargo.toml` to include:
`bevy_mod_picking = {git = "https://github.com/Fincap/bevy_mod_picking", branch = "migrate-bevy-0.10.0"}`

aevyrie (approved):
Looks great, thank you.

PR #214, opened 2023-06-11
Add a state resource for controlling the debug plugin

yyogo:
Implement runtime configuration for the debug plugin, as one of three states: `Disabled`, `Quiet` and `Noisy`. The `noisy` attribute is kept with a deprecation warning.


aevyrie (approved):

yyogo:
I renamed the 'Quiet' mode to 'Normal', I think that makes more sense.

aevyrie:
Thank you!

PR #230, opened 2023-07-10
Remove focuspolicy

aevyrie:
- Changed: The plugin no longer respects bevy_ui's `FocusPolicy`. This was proving to cause problems
  as mod_picking and bevy_ui have some fundamental differences that cannot be reconciled. This has
  been replaced by added fields on the `Pickable` component. You can use this to override the
  behavior of any entity in picking. This allows you to decide if that entity will block lower
  entities (on by default), and if that entity should emit events and be hover-able (on by default).
    - To make objects non-pickable, instead of removing the `Pickable` entity, use the new const
      value `Pickable::IGNORE`.
- Changed: The `Pickable` component is now optional. Backends can choose to use it for optimization,
  but should not filter out entities that do not have this component. An example of an optimization
  would be to early exit and stop raycasting once an entity is hit that blocks the entities below
  it.
- Changed: To fully remove bevy_ui as a dependency and avoid issues similar to the `FocusPolicy`
  change, bevy_mod_picking no longer updates the bevy_ui `Interaction` state. This has been replaced
  with a `PickInteraction` component that serves a similar purpose. This component aggregates the
  picking state of an entity (press, hover, none) across *all* pointers.
- Changed: `PickLayer`, used to order data from backends that targets the same render target, such
  as multiple render passes on the same window, has been changed from an `isize` to an `f32`. This
  change was made to support bevy_ui, which is "special" and can be rendered on any camera via a
  flag, instead of being rendered with its own camera. The bevy_ui backend now sets the order of any
  events emitted to be the camera order plus 0.5, which was not possible with an integer.
- Changed: The `PointerInteraction` component, which is added to pointers and tracks all entities
  being interacted with has changed internally from a hashmap of entities and their `Interaction` to
  a sorted list of entities and `HitData`, sorted by depth. This better reflects how pointer data
  would be expected to work, and makes it easier to find the topmost entity under each pointer.
- Added: `get_nearest_hit` method added to `PointerInteraction`
- Changed: Moved `PointerInteraction` from the `focus` module to `pointer`.


backends/bevy_picking_ui/src/lib.rs
```diff
@@ -135,21 +136,35 @@ pub fn ui_picking(
...
 
         while let Some(node) = iter.fetch_next() {
-            picks.push((
-                node.entity,
-                HitData {
-                    camera,
-                    depth,
-                    position: None,
-                    normal: None,
+            let mut push_hit = || {
```
bonsairobo, line 139: Ooh baby, baby! `push_it` real good!

crates/bevy_picking_core/src/focus.rs
```diff
@@ -134,11 +134,12 @@ fn build_hover_map(
                 .rev()
                 .flat_map(|depth_map| depth_map.values())
             {
-                pointer_entity_set.insert(entity, pick_data);
-                if let Ok(FocusPolicy::Block) = focus.get(entity) {
-                    break;
+                if pickable.get(entity).is_err() {
+                    error!();
```
bonsairobo, line 138: You forgot to say something.

crates/bevy_picking_core/src/focus.rs
```diff
@@ -134,11 +134,12 @@ fn build_hover_map(
                 .rev()
                 .flat_map(|depth_map| depth_map.values())
             {
-                pointer_entity_set.insert(entity, pick_data);
-                if let Ok(FocusPolicy::Block) = focus.get(entity) {
-                    break;
+                if pickable.get(entity).is_err() {
```
bonsairobo, line 137: You can `get` only once if you do:

```rust
match pickable.get(entity) {
    Ok(p) => {
        match p {
            Pickable::Block => {
                pointer_entity_set.insert(entity, pick_data);
                break;
            }
            Pickable::Pass => {
                pointer_entity_set.insert(entity, pick_data);
            },
            Pickable::Ignore => {}
        }
    }
    Err(e) => {
        error!("Pickable missing: {e}");
        continue;
    }
}
```

crates/bevy_picking_core/src/lib.rs
```diff
@@ -53,9 +52,19 @@ impl Default for PickingPluginsSettings {
     }
 }
 
-/// Used to mark entities that should be pickable.
-#[derive(Component, Debug, Default, Clone, Reflect)]
-pub struct Pickable;
+/// Used to mark entities that should be pickable. Similar to bevy's `FocusPolicy`, except this type
+/// allows entities to be completely ignored in the picking process, and does not require compiling
+/// bevy_ui.
```
bonsairobo, line 57: I think it's nice to just have the first sentence separated by a line break because rustdoc will truncate everything after the line break for the abbreviated text that appears next to doc items.

crates/bevy_picking_core/src/focus.rs
```diff
@@ -134,11 +134,12 @@ fn build_hover_map(
...
+                    error!();
+                    continue;
                 }
-                if focus.get(entity).is_err() {
+                pointer_entity_set.insert(entity, pick_data);
```
bonsairobo, line 141: As written, I think the `Ignore` case will still get inserted.

crates/bevy_picking_core/src/lib.rs
```diff
@@ -53,9 +52,19 @@ impl Default for PickingPluginsSettings {
...
+#[derive(Component, Debug, Default, Clone, Reflect, PartialEq, Eq)]
+pub enum Pickable {
+    /// This entity can be focused, but blocks lower entities from receiving picking focus.
+    #[default]
+    Block,
+    /// This entity can be focused, and also allows entities underneath this one to be focused.
+    Pass,
+    /// Ignore this entity when computing picking focus.
+    Ignore,
```
bonsairobo, line 66: I think you should clarify that `Ignore` blocks lower entities from focus.

backends/bevy_picking_ui/src/lib.rs
```diff
@@ -135,21 +136,35 @@ pub fn ui_picking(
...
+                ))
+            };
+            match node.pickable {
+                Some(pickable) => match pickable {
+                    Pickable::Block => {
+                        push_hit();
+                        break;
+                    }
+                    Pickable::Pass => push_hit(), // allow the next node to be hovered/clicked
+                    Pickable::Ignore => (),       // allow the next node to be hovered/clicked
```
bonsairobo, line 157: Shouldn't this `break`?

backends/bevy_picking_sprite/src/lib.rs
```diff
@@ -37,7 +36,7 @@ pub fn sprite_picking(
         &Handle<Image>,
         &GlobalTransform,
         &ComputedVisibility,
-        Option<&FocusPolicy>,
+        Option<&Pickable>,
```
bonsairobo, line 39: I'd make this mandatory.

crates/bevy_picking_core/src/focus.rs
```diff
@@ -134,11 +134,12 @@ fn build_hover_map(
...
+                if let Ok(Pickable::Block) = pickable.get(entity) {
```
bonsairobo, line 142: Should also `break` for `Ignore`.

aevyrie:
Had a discussion about this offline.

- Picking should work automatically, the `Pickable` component should only act as an override
- Add two booleans in the override:
  - `block_lower`: should this entity block interacting with lower entities
  - `emit_events`: should this entity emit events
  - By default, entities without the component should block and emit events.
- Backends can ignore this component, but may want to use it as a form of optimization. For example, during raycasting it might be desirable to early-exit if an entity is hit that blocks the entities below it.

PR #236, opened 2023-07-26
Re-export bevy_eventlistener

Newbytee:
Users of bevy_mod_picking may want direct access to bevy_eventlistener's exports when using certain features, such as what's done in the event_listeners.rs example. As such, re-export this module for easy use when using bevy_mod_picking. It also avoids having to keep bevy_mod_picking and bevy_eventlistener in sync in your local project.

I'm not familiar with how re-exports usually are used, so if this is unusual or bad somehow feel free to reject it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -172,6 +172,7 @@ use bevy::{prelude::Bundle, ui::Interaction};
 use bevy_picking_core::PointerCoreBundle;
 use prelude::*;
 
+pub use bevy_eventlistener::{self as eventlistener};
 pub use bevy_picking_core::{self as picking_core, backend, events, focus, pointer};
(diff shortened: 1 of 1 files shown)


aevyrie:
This is already available in the prelude: https://github.com/aevyrie/bevy_mod_picking/blob/36da26852e9ef71ae54cff7a3f740afb271f02d5/src/lib.rs#L214

Newbytee:
> This is already available in the prelude:
> 
> https://github.com/aevyrie/bevy_mod_picking/blob/36da26852e9ef71ae54cff7a3f740afb271f02d5/src/lib.rs#L214

What if I want to use something outside of `bevy_eventlistener`'s prelude?

PR #240, opened 2023-08-18
Add `Inside` to `Pointer` events

100-TomatoJuice:
Adds `Inside` to `Pointer`, which fires an event while the pointer is inside an entity. It is mostly the `Over` event, but instead of firing when it enters an entity, it just continues to fire.

### DIFF
--- a/crates/bevy_picking_core/src/events.rs
+++ b/crates/bevy_picking_core/src/events.rs
@@ -123,6 +123,14 @@ pub struct Move {
 }
 impl IsPointerEvent for Move {}
 
+/// Fires while a pointer is inside the `target` entity.
+#[derive(Clone, PartialEq, Debug, Reflect)]
+pub struct Inside {
+    /// Information about the picking intersection.
+    pub hit: HitData,
+}
+impl IsPointerEvent for Inside {}
+
 /// Fires when the `target` entity receives a pointer down event followed by a pointer move event.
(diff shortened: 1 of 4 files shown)


aevyrie:
This already exists as `Move`. 🙂

100-TomatoJuice:
@aevyrie Thanks for the quick response! I created `Inside` as if you start with the pointer already in an entity, `Move` doesn't resister. An example is I had an entity follow the cursor on a plane. But, the system could be turned off. So, if turned back on and the pointer is inside the entity, `Move` doesn't fire an event. Adding `Over` and `Move` to do one piece of a whole seems a little verbose. I feel like this for this use case `Inside` is helpful but, your thoughts about this use case would be insightful!

100-TomatoJuice:
Maybe an adjustment to how `Move` works would be better in this case though?

aevyrie:
I'm trying to keep the events limited to the events available on the web, simply to keep the available events easy to learn, and prevent them ballooning in quantity. The way the plugin is built, you can add your own `Pointer<Inside>` event and use it inside your own apps. You could even publish your own crate that adds this behavior on top of this plugin. See the selection crate for example.

100-TomatoJuice:
That makes sense. After thinking about it more, your take on this makes more sense as `Inside` offered very little and breaks away from web events. I did not know about adding custom pointer events, so thank you!

PR #252, opened 2023-09-10
cargo fmt

waywardmonkeys:
Newer Rust is now correctly formatting `let Some(..) = ...` constructs, so this is needed now for CI to pass.


aevyrie:
Thanks!

Issue #234, opened 2023-07-15
Support scale in sprite backend


NiklasEi:
The following sprite has a scale of 4 in x. The border is changed on `Over` and `Out` events.
https://github.com/aevyrie/bevy_mod_picking/assets/12236672/e74489ea-aad9-4e0f-9e10-2a1f9961f7eb

nametable:
I figured out how to add it for the basic case [here](https://github.com/nametable/bevy_mod_picking/tree/sprite-backend-transform-scaling). However, it does not perform as expected when there are child sprites of a scaled sprite. I might have to see how the renderer figures out the final scale

hmeine:
I was about to file a similar issue. In fact, I think rotation would need to be taken into account as well. In my case, sprites are scaled down by a factor of 1000, so the current behavior is that the effective hit box is 1000x too large and the sprite with the highest z coordinate wins.

Looking at the code by @nametable , I would think the solution is even much simpler: The existing GlobalTransform should be used instead of a new Transform query – that would be a smaller change, cheaper at runtime, and properly deal with nested transforms (child sprites as mentioned above).

hmeine:
After a night of sleep, I have another argument for why I would prefer being able to map the pointer position into the entity system (via an inverted GlobalTransform – did not see a way to access that publicly): That would make it possible to support checking the sprite image at that position for transparency. I guess that would be a cool option as well (could be gated through a Component, much like the unblocking currently).

And of course that would also deal with rotation.

hmeine:
I have opened two PRs implementing the above ideas. #255 (yay - 0xff) subsumes #254 and does the above-mentioned matrix inversion, supporting also rotated sprites.  I updated the example to demonstrate the result.

aevyrie:
This is now merged in main. Thanks for the issue and contribution!

PR #254, opened 2023-09-16
Fix hit box of scaled sprites

hmeine:
We want to take into account not only the translation part of the global transform, but also the scaling.  At the moment, rotations are still not dealt with.

I am also not sure how performant the `to_scale_rotation_translation()` method is.  (I think a different API could be faster, because we do not really need the decomposition.)

This should fix issue #234.


aevyrie (approved):

aevyrie:
Excellent, thank you!

PR #255, opened 2023-09-19
Support rotated sprites (and rotated hit boxes)

hmeine:
This is certainly more expensive than the old code, but also much more correct and future-proof: Inverting the affine transform allows to support rotated sprites as well.

I also adapted the `sprite` example to demonstrate the feature.

I am sure that glam offers very fast Affine3A inversion, yet I think it would be possible to speed this up in theory.  Not sure that caching the inverted transform makes much sense, since many global transforms would be invalidated very often.

On the other hand, we only perform a 2D hit box test, but the affine transform is 3D; one could skip computing the z coordinate AFAICS.

Finally, an option for speedups would be to make the kind of test configurable; if it is known that certain sprites are never rotated (or even scaled), the computation could take a simpler path.  Such configurability would also allow for a more expensive path: Testing the sprite pixmap for transparency at the pointer position would now be a low hanging fruit.

### DIFF
--- a/backends/bevy_picking_sprite/src/lib.rs
+++ b/backends/bevy_picking_sprite/src/lib.rs
@@ -75,19 +75,24 @@ pub fn sprite_picking(
                     if blocked || !visibility.is_visible() {
                         return None;
                     }
-                    let (scale, _rot, position) = sprite_transform.to_scale_rotation_translation();
+
+                    // hit box in sprite coordinate system
                     let extents = sprite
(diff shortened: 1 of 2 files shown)


aevyrie:
Thank you!

> Such configurability would also allow for a more expensive path: Testing the sprite pixmap for transparency at the pointer position would now be a low hanging fruit.

Definitely something I would like to support.

PR #260, opened 2023-09-30
Add example for changing the debug output state.

mattdm:
Created this as I struggled to understand it myself. Since someone came along with the same question I had, I figure an example might help others too.

I had a version where the button handling was more complicated, with one system to handle all button clicks. That version didn't require a second parameter for `.add_button`, but then ends up kind of complicated because the only distinguishing element is the `text` itself. Even with a simplified version from my initial try, I feel like it adds more complication that isn't really relevant to the example. If you want, though, I could take another look at that.


aevyrie:
Looks great, thanks!

PR #261, opened 2023-09-30
small suggestion for docs, giving pointers to bevy_eventlistener

mattdm:
This definitely would have helped me a few days ago. Maybe also others? Just a suggestion!


aevyrie (approved):
Nice addition, thanks!

Issue #259, opened 2023-09-29
Consider using debug!() and trace!() instead of info!() for logging


mattdm:
While the "info" level is clearly appropriate for bevy_mod_picking as a thing in itself, from the point of view of a higher-level consumer, that seems too escalated. I know it literally says "unofficial" in the title, but I think this guidance [from the Bevy Cheat Book](https://bevy-cheatbook.github.io/features/log.html#levels) makes sense:

>
>  * `error`: something happened that prevents things from working correctly
>  * `warn`: something unusual happened, but things can continue to work
>  * `info`: general informational messages
>  * `debug`: for development, messages about what your code is doing
>  * `trace`: for very verbose debug data, like dumping values

... and correspondingly I think it would be nice if this crate used `debug` for the "normal" messages and `trace` for the "noisy" ones. That leaves "info" available for more high-level messages.

mattdm:
Terminal screenshot to help illustrate why I think this would be useful. With this, the `info!` messages from the `bevy_ui` example stand out even with bevy_mod_picking debug messages also enabled.

![image](https://github.com/aevyrie/bevy_mod_picking/assets/123367/63d0938c-abb8-4879-9e4b-f7e079432d21)

aevyrie:
Really great feedback. I'm sold.

aevyrie:
Fixed by #262

PR #262, opened 2023-09-30
Change log level from info to debug and trace

mattdm:
This implements #259.

Messages linked to the "Noisy" debug state are logged with `trace!()`. The rest are logged with `debug!()`. Filtering on this _could_ replace the Noisy/Normal state distinction entirely (perhaps leaving the DebugPickingMode state _only_ for hiding or showing the cursor text overlay) but I'll leave that possibility for the future. This just makes a minimal change.


aevyrie (approved):
This is a much needed improvement. Thanks!

PR #263, opened 2023-10-01
configurable mapping from Mouse Buttons to Pointer Buttons

mattdm:
This adds a configurable map to InputPluginSettings. This allows for configuring Right, Left, and Middle buttons as determined by bevy, as well as Other buttons (by number). It does not extend picking to provide target mappings other than Primary, Secondary, and Middle. (That's future work!)

The abstraction here makes it easy for games to provide end-user configuration of the binding. I'm not set on _this_ particular interface to changing it and can update the code for other ideas.

If this is accepted, my next step is to tackle the "other" buttons.

### DIFF
--- a/crates/bevy_picking_input/src/lib.rs
+++ b/crates/bevy_picking_input/src/lib.rs
@@ -52,12 +52,16 @@ impl Plugin for InputPlugin {
 pub struct InputPluginSettings {
     run_mouse: bool,
     run_touch: bool,
+    /// Map buttons from Bevy's mouse input (e.g. `MouseButton::Left`) to Pointer
+    /// buttons (e.g. PointerButton::Primary).
+    pub mouse_button_mapping: mouse::MouseButtonMapping,
 }
 impl Default for InputPluginSettings {
     fn default() -> Self {
         Self {
             run_mouse: true,
             run_touch: true,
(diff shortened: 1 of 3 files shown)


mattdm:
Also I know from Discord chat that you weren't convinced that anyone would want this. If you'd prefer, thanks to your modular design, I can maintain this as a separate crate. But personally, I think it makes sense in general.

mattdm:
I'll update this this evening when I get a chance!

aevyrie:
I don't think this change is worth the added complexity because the input plugin already has this abstraction. A user can disable the mouse or touch inputs via the input settings:

https://github.com/aevyrie/bevy_mod_picking/blob/b66dfd3f859dba2d60d2265cf3064a5e432b2cfe/crates/bevy_picking_input/src/lib.rs#L62-L74

and send their own `InputPress` events using whatever logic they require. The reason I created the `InputPress` event was to decouple inputs from hardware, and to allow users to drive pointer presses with whatever their hearts desired.

I appreciate the contribution, but as it stands I'm pretty hesitant to add more complexity to an already large plugin that aspires to be upstreamed. 🙂

aevyrie:
To your comment about the crate, I think that's a great idea. If you want to create a fully featured mouse input plugin, that would be super cool! I'm mostly concerned about managing complexity inside this plugin.

PR #266, opened 2023-10-03
Fix ordering bug in SelectionPlugin

TheRawMeatball:
`send_selection_events` and `update_state_from_events` not being ordered relative to one another meant that sometimes clicks would be applied to `PickSelection` one frame late, and sometimes not. This led to unreliable and sometimes broken behavior, so we fix their ordering using a `.chain()` call to ensure `PickSelection` is always updated on the frame the pointer is released.

### DIFF
--- a/crates/bevy_picking_selection/src/lib.rs
+++ b/crates/bevy_picking_selection/src/lib.rs
@@ -231,7 +231,6 @@ pub fn update_state_from_events(
 ) {
     for selection in selections.iter() {
         if let Ok(mut select_me) = selectables.get_mut(selection.target) {
-            dbg!("mod_picking");
             select_me.is_selected = true;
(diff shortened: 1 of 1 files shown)


crates/bevy_picking_selection/src/lib.rs
```diff
@@ -229,6 +231,7 @@ pub fn update_state_from_events(
 ) {
     for selection in selections.iter() {
         if let Ok(mut select_me) = selectables.get_mut(selection.target) {
+            dbg!("mod_picking");
```
aevyrie, line 234: This probably shouldn't be here.

aevyrie (changes requested):

aevyrie (approved):

PR #285, opened 2023-11-27
Fix the Bevy UI backend so it ignores clipped areas of UI nodes

ickshonpe:
The Bevy UI backend doesn't check for clipping, and allows interaction with clipped non-visible sections of UI nodes.
 
This PR intersects the node rect with its calculated clip rect (if it has one) and then checks if the clipped node contains the cursor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # UNRELEASED
 
 - Added: support for `bevy_ui` `UiScale`.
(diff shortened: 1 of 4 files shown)


aevyrie (approved):
Looks good, but needs a clippy fix before this can merge. Thank you!

ickshonpe:
> Looks good, but needs a clippy fix before this can merge. Thank you!

Should be fixed now

PR #289, opened 2023-12-14
Add the `RayMap` resource and use it in rapier and mod_raycast backends

bonsairobo:
The goal is to make it easier to build raycasting backends and to make it possible to easily get a `Ray` while building interactive entities that handle picking events.


bonsairobo:
BTW @aevyrie, might be worth adding this to the changelog:

```
- Fixed: Ray construction now respects DPI / window scale
```

Ideally there would also be an example that shows this is fixed. Up to you if you want that to happen in this PR or later. I can help if you like.

crates/bevy_picking_core/src/backend.rs
```diff
@@ -107,3 +113,114 @@ impl HitData {
         }
     }
 }
+
+pub mod ray {
```
bonsairobo, line 117: You have something against file modules?

bonsairobo: Oh I'm guessing you just didn't want to create a subdirectory for one module.

PR #293, opened 2023-12-27
Account for near clipping plane in sprite backend

teodosin:
Depth is now calculated as distance of the entity to the camera's near clipping plane, to be consistent with the raycast backend.


aevyrie:
Looking at this a bit more closely, none of the other backends use depth in the traditional sense, rather, they use distance.

PR #296, opened 2023-12-31
replace all .single calls with matched .get_single calls

Schmarni-Dev:
replace all .single calls with matched .get_single calls to avoid Crashing in environments where there is no window available


aevyrie (commented):
Can you please update the changelog?

PR #304, opened 2024-01-15
added filtering of 3d and 2d cameras to allow for 3d and 2d cameras to operate simultaneously

skylerparr:
When rendering with 2D and 3D cameras, the results would be very unpredictable.  Adding the filter fixed the issue.

### DIFF
--- a/backends/bevy_picking_raycast/src/lib.rs
+++ b/backends/bevy_picking_raycast/src/lib.rs
@@ -15,6 +15,7 @@
 #![deny(missing_docs)]
 
 use bevy_app::prelude::*;
+use bevy_core_pipeline::prelude::Camera3d;
 use bevy_ecs::prelude::*;
 use bevy_reflect::prelude::*;
 use bevy_render::{prelude::*, view::RenderLayers};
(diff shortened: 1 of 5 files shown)


backends/bevy_picking_raycast/src/lib.rs
```diff
@@ -66,13 +67,16 @@ pub fn update_hits(
...
+    picking_cameras: Query<
+        (
+            Entity,
+            &Camera,
+            &GlobalTransform,
+            Option<&RaycastPickable>,
+            Option<&RenderLayers>,
+        ),
+        With<Camera3d>,
+    >,
```
Azorlogh, line 79: I think this will prevent 2d cameras from picking 2d meshes

aevyrie:
The raycasting backend should be reporting the camera.order, which the picking plugin uses to sort hits. Do you have a minimal repro of the "unpredictable picking" bug with 2d/3d cams?

aevyrie:
It appears this is already fixed on `main` thanks to #293, which added `OrthographicProjection` to the sprite backend query. I've modified the existing `sprite` example with a 3d camera so we can test for this in the future.

skylerparr:
Thank you. Sorry for the radio silence. I was dealing with a sick baby. I appreciate this library.

PR #333, opened 2024-05-24
derive `Copy` for `HitData` and adjust consumers

databasedav:


databasedav:
from discord https://discord.com/channels/691052431525675048/1038322714320052304/1243630594928279704
> Copy doesn't change the perf of those structs, but it will make it a breaking change to add a non-copy type to them in the future.
Also want to be careful implementing it on types that are mutated, to prevent copying when moving is desired.
That can lead to the situation where you think you have mutated something, but you've actually just mutated a copy, and the original is unchanged.

PR #354, opened 2024-07-23
Sprite backend improvements

jnhyatt:
Sprite backend was not reporting world position or normal for hits. Upon further investigation I found that the backend was also not correctly hit testing sprites if they had rotation about any axis other than the Z axis. This was producing some very surprising results in some cases (i.e. reported hits weren't even close to the sprites).

This pr fixes both of these issues: sprite hits are now determined by intersecting the sprite's bounding rect with a 3d line segment extending from `(cursor_x, cursor_y, near_clip)` and `(cursor_x, cursor_y, far_clip)`. This should return correct results for any arbitrary sprite transform. Hit position and normal are now being populated as well.

One question I have about the impl is whether backends are supposed to return hits in any particular order. Before, depth was being determined by sprite's Z coordinate, but since sprites now support arbitrary rotations, it's possible that a sprite with a deeper Z value could be picked in front of another. Does that mean that the hits should be sorted after hits are collected? And if so, does that mean that the sprite sorting that happens before the hit-testing can/should be removed?

### DIFF
--- a/backends/bevy_picking_sprite/src/lib.rs
+++ b/backends/bevy_picking_sprite/src/lib.rs
@@ -81,10 +81,12 @@ pub fn sprite_picking(
             continue;
         };
 
-        let Some(cursor_pos_world) = camera.viewport_to_world_2d(cam_transform, location.position)
+        let Some(cursor_ray_world) = camera.viewport_to_world(cam_transform, location.position)
         else {
             continue;
         };
+        let cursor_ray_len = cam_ortho.far - cam_ortho.near;
+        let cursor_ray_end = cursor_ray_world.origin + cursor_ray_world.direction * cursor_ray_len;
 
         let picks: Vec<(Entity, HitData)> = sorted_sprites
             .iter()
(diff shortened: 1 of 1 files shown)


aevyrie:
The answers to your questions are documented here: https://docs.rs/bevy_mod_picking/latest/bevy_mod_picking/backend/struct.PointerHits.html#structfield.picks

> An unordered collection of entities and their distance (depth) from the cursor.

and here: https://docs.rs/bevy_mod_picking/latest/bevy_mod_picking/backend/struct.HitData.html#structfield.depth.html

>  depth only needs to be self-consistent with other [PointerHits](https://docs.rs/bevy_mod_picking/latest/bevy_mod_picking/backend/struct.PointerHits.html)s using the same [RenderTarget](https://docs.rs/bevy_render/0.14.0/x86_64-unknown-linux-gnu/bevy_render/camera/camera/enum.RenderTarget.html). However, it is recommended to use the distance from the pointer to the hit, measured from the near plane of the camera, to the point, in world space.

PR #365, opened 2024-09-04
fix: Account for viewport position in sprite backend

### LINKED ISSUE #364: Sprite picking is offset when `Camera::viewport` is set
if the viewport is not in the top left, the picking is offset by the distance to the top-left (I think). At least it's incorrect.

johanhelsing:
Fixes #364


andriyDev:
This has been merged into Bevy as https://github.com/bevyengine/bevy/pull/16386

